PR #162, opened 2014-06-25
Fix Azure Disconnect problem.

alfa-jpn:
I'm poor at English.
If you can't understand what I mean, please let me know.

---

Azure SQL close the client connection every 5min.
FreeTDS can't diagnose it.

This is a specification of Azure SQL.
http://blogs.msdn.com/b/bartr/archive/2010/06/18/sql-azure-connection-retry.aspx

Adapter needs execute SQL.
But, this method is too slow, because client waits timeout.
That's way this pull request measure the time (5min) and execute SQL.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,7 @@ misc
 *.gem
 ports
 rake-compiler-dev-box/
+
+# IDE
+*.iml
+.idea/
--- a/lib/tiny_tds/client.rb
+++ b/lib/tiny_tds/client.rb
@@ -74,6 +74,11 @@ module TinyTds
       opts[:encoding] = (opts[:encoding].nil? || opts[:encoding].downcase == 'utf8') ? 'UTF-8' : opts[:encoding].upcase
       opts[:port] ||= 1433
       opts[:dataserver] = "#{opts[:host]}:#{opts[:port]}" if opts[:dataserver].to_s.empty?
+
+      @azure = opts[:azure]
+      @azure_timeout = opts[:azure_timeout] || 300
+      @azure_timeout_at = Time.now + @azure_timeout
+
       connect(opts)
     end
     
@@ -83,7 +88,14 @@ module TinyTds
     end
 
     def active?
-      !closed? && !dead?
+      return !closed? && !dead? unless @azure
+      return false if @azure_timeout_at < Time.now
+
+      @azure_timeout_at = Time.now + @azure_timeout
+      execute('SELECT 1').each
+      true
+    rescue TinyTds::Error
+      false
     end
 
     private


jwoertink:
I have this same issue, and I'm use the fork currently. It *seems* to be working from what I can tell. I'm still having delays, but they're nowhere near as bad as they were.

metaskills:
FYI, I have setup my own Azure account and will be testing this and a few others things.

metaskills:
I did confirm that waiting on a connection for 300 seconds will get a timeout/disconnect on the next query. However, this [MSDN article](https://msdn.microsoft.com/en-us/library/ee336245.aspx#cc) suggests that connections could be timed out after 30 minutes vs the 5 I just experienced. Perhaps the duration is variable and the 5 suggested here would be too low for heavy users of Azure?

jwoertink:
Ah interesting. Well, in my case, my app isn't used very often. It's basically a kiosk. I have a javascript timeout that kicks off so if no one has touched the kiosk in several minutes, it will automatically redirect to the home page of the app. On the home page, it makes a query to check for certain details, and in the case of the azure timeout, it fails on the redirect because the app sat dormant for over 5 minutes. It could sit dormant for a full weekend at times though, so for me, even a 30min timeout wouldn't be enough. 

Maybe there's an option like..

```yml
azure: true
azure_timeout: 0 # no timeout
```
or
```yml
azure: true
azure_timeout: 30 # 30minute timeout
``` 
and if not configured, then it's defaulted to 5?

metaskills:
Yea, a config could make sense. We use `nil` in Ruby too :) 

I will be thinking about this some more too since I will like want to re-enchance [this method](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/blob/master/lib/active_record/connection_adapters/sqlserver_adapter.rb#L132-L138) in the adapter too to delegate once again to TinyTDS's `active?` method.

metaskills:
Based on a huge amount of research I did for the adapter issue (https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/issues/402) I think I found a more appropriate solution. This [comment in that thread](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/issues/402#issuecomment-107074580) talks about the `reaping_frequency` ActiveRecord config. Can you expierment by setting this to something like 29 minutes? This config takes a seconds value. So use `1740` for that. More info:

* [The Rails Grim Reaper](http://omegadelta.net/2014/03/15/the-rails-grim-reaper/)

metaskills:
OK, after another round of discussions with both Microsoft and Rails core team members about connection management. I am almost certain this is something that TinyTDS should not address. Connection management should be the responsibility of connection pool gems and such. So in Rails this is handled for us using the built in middleware and Rails interfaces. Please read this thread again for full details.

https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/issues/402#issuecomment-107074580

All we need to do at the TinyTDS layer is a simple #active? method.

jwoertink:
awesome! Thanks for the heads up.

larskanis:
@metaskills Yes, this is how we do it in the pg.gem as well.

PR #174, opened 2014-11-04
Make first work like expected

middagj:
It took me quite a while to get that first wasn't working like expected for queries with multiple resutls. I got a
`Attempt to initiate a new Adaptive Server operation with results pending (TinyTds::Error)`
Probably I should do a `cancel` after, but than I have to save the result in a variable which makes it less expressive.

Next to that I found out that you could not run `each` after you ran `first`. To make the Result object to behave more like an Enumerable I propose this change which also improves performance compared to what `first` does now.

### DIFF
--- a/lib/tiny_tds/result.rb
+++ b/lib/tiny_tds/result.rb
@@ -3,6 +3,8 @@ module TinyTds
     
     include Enumerable
     
-    
+    def first
+      each(first: true).first
+    end
   end
 end
--- a/test/result_test.rb
+++ b/test/result_test.rb
@@ -45,6 +45,13 @@ class ResultTest < TinyTds::TestCase
       assert_equal data.first.object_id, result.each.first.object_id
     end
     
+    it 'allows successive call to each after first' do
+      result = @client.execute(@query1)
+      first = result.first
+      data = result.each
+      assert_equal first.object_id, data.first.object_id
+    end
+    
     it 'returns hashes with string keys' do
       result = @client.execute(@query1)
       row = result.each(:as => :hash, :symbolize_keys => false).first


middagj:
No the first method (and probably the other Enumerable methods) calls the each method with a block. Amending the each method to return a result instance if not called with any arguments should not matter, although it is not very intuitive.

But I think you that each and the Enumerable methods do not work as expected for the result object as in that if each does not retrieve the full set of data you have to do a manual cancel. It stops me from writing
```ruby
return conn.execute(sql).first
```
and instead I have to write
```ruby
result = conn.execute(sql)
row = result.first
result.cancel
return row
```
now I have in my code
```ruby
return conn.execute(sql).each(first: true).first
```
I think the cancel burden should not be on the caller. You are correct the my patch only fix it for first, but I thought it would be a common use case, especially since you can pass first as argument in each. Do we actually need the full Enumerable interface? We can also supply the few methods that are actually used.

metaskills:
Hey Jacob, thanks for stopping by again.

> and instead I have to write...

Verbose I know, but it depends on your needs. When in ActiveRecord, with TinyTDS, you just use `connection.execute(sql)` and it works as expected. But for TinyTDS, I think we want to keep the implementation close to the C interface of DBLIB which we are building on top of. It is too easy to write a small wrapper around that.

> I think the cancel burden should not be on the caller.

But that is one of the cool things. When calling each, you are literally pulling data from the wire at that same time. Again, I wanted to sticks to close to the DBLIB interface.

> Do we actually need the full Enumerable interface

Maybe not. And maybe we should yank Enumerable inclusion. In fact, maybe the first interface to each should go as well. I would be more in favor of (a) finding the place where we can make Result more Enum compliant or (b) yanking in favor of something else. My hope was that we could do (a) by fixing some small oversight of mine in the `rb_tinytds_result_each` implementation, but I am not seeing it yet and was hoping for more feedback.

middagj:
Hi Ken,

I can understand the intention to stick close to the C library. It makes it easy to use your own abstraction or wrapper. I prefer not to use the ActiveRecord library when making small scripts against our database server. And I like the fact that the library reflects the actual networking activity.

But calling each the second time, does not pull data off the wire, but instead of the cached result, i.e. the second `!NIL_P(block)`. That makes specifying parameters like `first` only valid for the first time.

If we could make `each` iterate over a new data set on every invocation we could use an `ensure` block to make sure that the result set is cancelled when the user is done with the invocation. That would make the `each` method more easily Enum compliant and reduces the annoying `Attempt to initiate a new Adaptive Server operation with results pending (TinyTds::Error)` errors.

metaskills:
@middagj That could be good advice. I am going to close this issue but it does not mean the conversation has to stop. As I re-dig into the implementation of the each method and the way I have allowed it to take query options, specifically `cache_rows` and `as`, these refactors may be at odds. I even just tried to change `rb_tinytds_result_each` so that it would return the result object if rows were not cached. No luck.

I'd like to remove the hint that `TinyTds::Result` mixes in or acts like an `Enumerable` at all. But for some reason the `returns bigint for #insert when needed` test failed and caused the Client#insert to kill the connection.

middagj:
It is the [detect method](http://ruby-doc.org/core-2.2.1/Enumerable.html#method-i-detect) of `Enumerable` that is causing the fail.

Issue #200, opened 2015-05-26
Using TinyTDS for SQL Azure on Windows


meet-bhagdev:
I am struggling to make TinyTDS work for SQL Azure on Windows. I am aware that I have to configure and build FreeTDS using OpenSSL,I followed all the required steps and it still did not work. Is there anyway I can get TinyTDS(included FreeTDS) available for direct install?
Thanks!

metaskills:
Hey Meet! Please read issue #196 where @aharpervc started working on getting Windows binaries published by fixing rake-compiler-dev-box. So my question is, does the latest rake-compiler-dev-box have OpenSSL and can we change our port file to take advantage of this? If so, this is what issue https://github.com/rails-sqlserver/tiny_tds/issues/75 is for too.

meet-bhagdev:
Hi metaskills, no luck yet :( Still trying to make TinyTDS work on windows with SQL Azure. Got it to work with SQL Azure on Mac though so thats good. But if someone can help with the windows installations that would be great. gem install tiny_tds doesn't do the trick unfortunately.

metaskills:
Gonna close this issue while we work on #207 and get another pre-release of native windows binaries. That issue also adds OpenSSL too which is needed for Azure. Please watch that issue for updates. Cheers

Issue #196, opened 2015-05-08
Ruby 2.2 & 0.6.3 Release


jalmargyyk:
It seems that ruby 2.2 compatibility is only in master. Any chance for a new release?

metaskills:
It is not yet easy for me to do the precompiled Windows binaries. If you are on a POSIX system, you can just bundle to master.

jalmargyyk:
Ah, sorry I didn't think other platforms then linux. I prefer to use released gems as this is used in production. Well, I'll just use previous ruby version for the time being. Thanks.

aharpervc:
Referencing master is kind of a hassle when writing a gem (cant reference the repo in a gemspec). And it would be great to have an official release for 2.2. Is there any technical hold up?

metaskills:
> Referencing master is kind of a hassle when writing a gem 

Great points! 

> Is there any technical hold up?

Getting the latest master branch to work with [rake-compiler-dev-box](https://github.com/rails-sqlserver/tiny_tds#compiling-gems-for-windows) for the Windows binaries. Anyone got some free time to debug? I suspect I could make a release for just any system that can compile natively too.

metaskills:
I have pushed 0.6.3.rc1 which is POSIX only. No pre-build win binaries now. HTH!

aharpervc:
A windows build is also necessary. I have attempted to get rake-compiler working for ruby 2.2, not yet successful: https://github.com/rake-compiler/rake-compiler/issues/112

aharpervc:
Good news, I have successfully cross compiled tiny_tds 0.6.3.rc1 for ruby 2.2, for windows. And it works! So getting 0.6.3 released for all platforms for all supported ruby versions for real should totally be possible.

I believe there are only two major blocking issues to make this work. These might be rake-compiler issues or things that can be patched from this repo, not sure. I'm posting my progress here as I work on solutions in case anyone has some quick fixes.

In order to cross compile a gem for ruby 2.2, rake-compiler needs to cross build ruby 2.2. System ruby >= 2.0 is needed to build 2.2, but that's nbd since rake-compiler already acquires ruby 2.0 and 2.1. Rake-compile-dev-box has a prep script to cross build target ruby versions. [Here's the patch to add ruby 2.2](https://gist.github.com/aharpervc/65b3588f5225a118e7a8).

[...]

Issue #75, opened 2012-03-22
OpenSSL With RakeCompiler & Miniportile


metaskills:

metaskills:
Closing this issue since #207 will address it.

PR #207, opened 2015-07-16
Refactor build of freetds and iconv recipes and add OpenSSL.

larskanis:
extconf.rb now checks for availability of freetds, with a fallback to the bundled mini_portile recipes.

This allows native build on Windows with download and compilation of all dependent libraries. This also adds a appveyor config for ci-testing of the native build. However [not all tests pass, so far](https://ci.appveyor.com/project/larskanis/tiny-tds). Unfortunately I don't have the experience with MSSQL to fix this.

For building windows binary gems this makes the switch to the new rake-compiler-dock. As described in the README, Windows binary gems can be built per `rake gem:windows`. I tested the build on Windows and Linux and the resulting gems on Windows 7 with Ruby-2.2.2-x86 and Ruby-2.2.2-x64.

The mini_portile recipes are changed, now, to build shared libraries. This decreases the fat binary Windows gem from around 7 MB to 2.5 MB, compared to a static build. To avoid any version mismatches with OpenSSL, I used a more specific naming scheme for OpenSSL dlls. This should avoid any contact with the DLL hell.

With shared libraries checking for iconv should not be necessary. That's why I removed it.

This fixes issues #200, #75 and hopefully helps for #196.

EDIT:
As a further extension, I added the freetds source tar to the gem file. This should help for offline installations.

If someone would like to try the resulting gems, I've uploaded they [here](https://github.com/larskanis/tiny_tds/releases/tag/0.6.3.rc1.lars1) .


metaskills:
@aharpervc Can you take a look at this? Does it jibe with the latest rake-compiler-dev-box work you have done in master?

larskanis:
It actually removes the rake-compiler-dev-box things from the Rakefile (although it should still build with it). However I recommend to switch to rake-compiler-dock. It is now under the "official" rake-compiler organisation and has several advantages over rake-compiler-dev-box. [concurrent-ruby](https://github.com/ruby-concurrency/concurrent-ruby) has done this step, for instance, although they had a even more [extended version of rake-compiler-dev-box](https://github.com/ruby-concurrency/rake-compiler-dev-box).

aharpervc:
A few thoughts

* I'm definitely :+1: on setting up AppVeyor
* I'm :-1: on switching to something that's based on boot2docker, because instead of using vagrant as RCDB does, it's hardcoded to use virtualbox for some reason. I personally don't (and can't) use virtualbox, meaning with this PR I personally can no longer build this gem.
* I don't exactly understand why your mini_portile recipe changes change the size of the fat gems for you. When I build with RCDB they're only ~2.5MB. Perhaps I've misconfigured something and am actually relying on a shared lib on this machine? Not sure. Philosophically I think I disagree with moving away from a static build, so :-1: on that.
* :+1: on including dependency sources in one of our repos. I also noticed this with RCDB when sourceforge was down last week, since the mingw compilers are downloaded on the fly.

larskanis:
> I'm :-1: on switching to something that's based on boot2docker, because on Windows instead of using vagrant as RCDB does, it's hardcoded to use virtualbox for some reason.

Not really. It depends only on docker and as the most common solution for docker on Windows and OS-X, it has some direct support for boot2docker in addition. However it works on Vagrant, equally. You're using rake-compiler-dev-box on Hyper-V? If I understand correctly, Windows-8.1 forces to use Hyper-V. Could it be a solution, that rake-compiler-dock supports Vagrant and therefore Hyper-V, equally to boot2docker?

> I don't exactly understand why your mini_portile recipe changes change the size of the fat gems for you. When I build with RCDB they're only ~2.5MB.

This is without OpenSSL. I first linked everything statically and got gems with more than 7 MB. This is because freetds, OpenSSL and iconv are stored 3 or 4 times within the same gem. Therefore I switched to using shared libraries for freetds, OpenSSL and iconv. Gems are then 2.1 and 2.4 MB in size.

aharpervc:
> Not really. It depends only on docker and as the most common solution for docker on Windows and OS-X, it has some direct support for boot2docker in addition. 

Running `rake gem:windows` results in

> Docker is not available.
Please download and install boot2docker:
    https://github.com/boot2docker/windows-installer/releases

That message is correct; docker is not available and I did not have boot2docker installed. However, as I said, boot2docker has a hardcoded dependency on virtualbox which makes it non-viable for me.

> However it works on Vagrant, equally. You're using rake-compiler-dev-box on Hyper-V? If I understand correctly, Windows-8.1 forces to use Hyper-V. Could it be a solution, that rake-compiler-dock supports Vagrant and therefore Hyper-V, equally to boot2docker?

I have been using RCDB with hyper-v. I assume boot2docker's docker-ready distro would work in hyper-v (but I haven't verified this yet). In fact, vagrant has a docker provider which is backed by the boot2docker distro so presumably `vagrant up --provider=docker` would work on any system to bring up a docker based build environment. I'm not sure how to tell rake-compiler-dock to do this, though.

(side note: this should be pointing at the `rake-compiler` repo now right? https://github.com/rake-compiler/rake-compiler-dock/blob/master/Rakefile#L5)

> This is without OpenSSL. I first linked everything statically and got gems with more than 7 MB. This is because freetds, OpenSSL and iconv are stored 3 or 4 times within the same gem. Therefore I switched to using shared libraries for freetds, OpenSSL and iconv. Gems are then 2.1 and 2.4 MB in size.

Good to know. Clarify this point for me: building tiny_tds with shared libraries for those 3 dependencies, anyone who does `gem install tiny_tds` must previously have these libs on their system, correct? I had thought that we would avoid that separate dependency management with a (larger) static build that included those deps. Is that correct? If so, I would strongly recommend that we sacrifice a few MB to avoid forcing people to track down & build compatible deps.

aharpervc:
Also why did you switch `mini_portile` from a dev dependency to a runtime dependency @larskanis ?

wbond:
@aharpervc in terms of the file size, I believe @larskanis is referring to the fact that with static linking, each version of tiny_tds for the different versions of ruby have OpenSSL and iconv linked in. With dynamic linking, there is one set of shared libraries that all of the different versions of ruby can depend on.

I would imagine ideally the best solution would be one release for each version of ruby instead of one release for all versions of ruby, but that may add more development overhead, whereas bandwidth is someone else's concern.

aharpervc:
> In fact, vagrant has a docker provider which is backed by the boot2docker distro so presumably vagrant up --provider=docker would work on any system to bring up a docker based build environment

Welp, not so fast. [hashicorp's boot2docker vagrant images only support vmware and virtualbox](https://atlas.hashicorp.com/hashicorp/boxes/boot2docker)

larskanis:
> building tiny_tds with shared libraries for those 3 dependencies, anyone who does gem install tiny_tds must previously have these libs on their system, correct?

Definitely not! All dependent DLLs are bundled into the fat binary gem, of course. No different use to a static linked version.

> Also why did you switch mini_portile from a dev dependency to a runtime dependency @larskanis ?

There are at least two reasons to make mini_portile a runtime dependency and do the build in extconf.rb:

1. We can build the source gem native on Windows, without the requirement to install any dependent libraries on the system before (per chocolatey etc.). It checks for availability of these libraries, and uses these, but if the requirements are not meet, it builds freetds, iconv and openssl on it's own per DevKit. I forgot to mention the native build on Windows, but that's what is [obviously required on appveyor](https://ci.appveyor.com/project/larskanis/tiny-tds).

2. Libraries are not mixed between different ruby and compiler versions. This is often the case, when building per rake tasks. rake-compiler allows to set different configure options depending on the architecture (x86/x64) but not on the ruby version. Therefore current versions of tiny_tds uses the libraries of the old gcc-4.2.1 for all x86 ruby versions, although ruby-2.0 jumped to DevKit with gcc-4.7.2.

> I would imagine ideally the best solution would be one release for each version of ruby instead of one release for all versions of ruby, but that may add more development overhead, whereas bandwidth is someone else's concern.

Rubygems doesn't support this.

> Welp, not so fast. hashicorp's boot2docker vagrant images only support vmware and virtualbox

There are several ways to use Vagrant and docker. I'll try some ways out.

@aharpervc Do you know, how Vagrant maps the shared folder `/vagrant` into the box? As far as I know Hyper-V does not support shared folders like Virtualbox. What shows `mount` in a Hyper-V rake-compiler-dev-box?

wbond:
That's unfortunate that rubygems doesn't support per-ruby release packages. Oh well. (I don't actually work in Ruby any more, just worked on this at one point.)

larskanis:
I noticed, that the appveyor link doesn't lead to the right test. any longer, but to the test of #208 . The following link looks much better: https://ci.appveyor.com/project/larskanis/tiny-tds/build/1.0.14

aharpervc:
> explanations

Ok, makes sense I guess. I suppose I retract my philosophical "-1" on the shared library issue

> There are several ways to use Vagrant and docker. I'll try some ways out.

Agree... I was already fiddling around to see how difficult it would be to simply create a new/forked vagrant box configuration with boot2docker hyperv. Not sure yet what it'd take...

> Do you know, how Vagrant maps the shared folder /vagrant into the box? As far as I know Hyper-V does not support shared folders like Virtualbox. What shows mount in a Hyper-V rake-compiler-dev-box?

It appears to work the same as other providers, mapping the working dir to `/vagrant`. I believe it creates a windows share to do this. `mount` reports `//192.168.250.97/798459085011fffe0eddba964f4c5f33 on /vagrant type cifs (rw)` (among others)

larskanis:
> mount reports //192.168.250.97/798459085011fffe0eddba964f4c5f33 on /vagrant type cifs (rw)

Makes sense, that's what I expected. However Vagrant on Virtualbox makes use of Virtualbox specific shared folders.

metaskills:
@larskanis Thank you so much for this pull request! Just getting to reading it today in full.

>  Philosophically I think I disagree with moving away from a static build

As I read thru this, that is my first reaction as well.

> Definitely not! All dependent DLLs are bundled into the fat binary gem, of course. No different use to a static linked version.

Wow! Really? How do you build smaller "dynamic" gems that have DLLs inside and act static? 

> switch mini_portile from a dev dependency to a runtime dependency

Whoa, those reasons sound wicket. I will take a closer look too.

> Re AppVeyor

I love the idea of using this. I will set something up under the `rails-sqlserver` account and links will use this once we nail down other thing.

metaskills:
I am having a hard time checking this PR out locally. Using what GitHub recommended

```
git checkout -b larskanis-rake-compiler-dock master
git pull git://github.com/larskanis/tiny_tds.git rake-compiler-dock
```

Can you help with this? Maybe rebase to latest master?

larskanis:
It's rebased to master. #202 was already in.

larskanis:
ftp.freetds.org seems to be down, so [appveyor-tests fail](https://ci.appveyor.com/project/larskanis/tiny-tds/build/1.0.16) to build the extension for the rebased version :cry:

metaskills:
Lars, this is amazing work. Most of it is way above my head, but it will be pulled in. The ftp.freetds.org being down is killing me now! Argh! As soon as it is back up I will do a bit more testing and and release `.rc2` gems. Great work and thank you so much!!!

larskanis:
> Most of it is way above my head, but it will be pulled in.

I don't think it's too complicated.  I tried to comment the least obviously
lines of code, but please ask, if there are unclear things or objections.

Mini_portile will probably change the way how configure_options are
interpreted in version 0.7.0. In this case, some quotes are distracting and
must be removed. I'll open a pull request, then (hoping I don't forget
about it).

And thank you for all your efforts in developing and maintaining this great
library and it's rails integration!

larskanis:
If someone would like to try the resulting gems, I've uploaded they [here](https://github.com/larskanis/tiny_tds/releases/tag/0.6.3.rc1.lars1) .

metaskills:
OK, doing a test of these now. Looking over your two new commits too @larskanis

metaskills:
I've merged this in. I will be doing some tests to setup appveyor under the rails-sqlserver account too.

metaskills:
Will report back on gem releases later today too.

meet-bhagdev:
As per https://github.com/rails-sqlserver/tiny_tds/issues/214, this seems to be complete :) :+1: 
Great job guys!

dharmdip:
Hi @metaskills,

i am getting this error while connecting, any thought?

tiny_tds-0.6.3.rc2-x64-mingw32/lib/tiny_tds/client.rb:74:inconnect': Write to the server failed (TinyTds::Error)

Thanks, 
Dharmdip

metaskills:
No immediate thoughts, can you provide more information like SQL Server version, authentication technique, connection settings, etc? Have you tried alternate login methods and making sure your SQL Server has TCP/IP connections on?

dharmdip:
Hi @metaskills, 

I am using with AR, and this is connection string. 

development:
  adapter: sqlserver
  mode: dblib
  dataserver: xxxxxxxx.database.windows.net'
  database: database
  username: username
  password: password
  timeout: 5000
  azure: true
  port: 1433

I am trying with FreeTds, any thought?

metaskills:
@dharmdip Please read this section of the readme. 

https://github.com/rails-sqlserver/tiny_tds#using-tinytds-with-azure

It outlines how to connect to Azure WRT the username which is incorrect above. I also see that you should remove the following options for your DB yaml as they are moot for your use case.

* mode - DBLIB is the default, you do not need to be explicit.
* port - Ports are kind of ignored for Azure. Let the `azure: true` do the driving. 1433 is also the default. Again, no need to be explicit.
* database - There is no `USE` with Azure. Let it do the work.

dharmdip:
Thank you @metaskills, 

Let me check.

dharmdip:
Stuck with same error, 

development:
  adapter: sqlserver
  dataserver: xxxxx.database.windows.net
  database: database_name
  username: username@xxxxx
  password: password  
  azure: true

metaskills:
Try removing the `xxxxx` from the `dataserver` option. Seems redundant when it is part of the username.

dharmdip:
okay, something new. 

/ruby/lib/ruby/gems/2.0.0/gems/tiny_tds-0.6.3.rc2-x64-mingw32/lib/tiny_tds/client.rb:74:in `connect': Unable to connect: Adaptive Server is unavailable or does not exist (TinyTds::Error)

dharmdip:
is this firewall or network issue?

meet-bhagdev:
Hello @dharmdip: Please check out the following for connection to Azure SQL DB using TinyTDS from 
1. [Windows](https://github.com/meet-bhagdev/test/blob/master/Ruby-Windows-SQL-DB.md)
2. [Mac](https://azure.microsoft.com/en-us/documentation/articles/sql-database-develop-ruby-simple-mac-osx/)
3. [Linux](https://azure.microsoft.com/en-us/documentation/articles/sql-database-develop-python-simple-windows/)

I have tried and tested the driver on all three OS's and it works just fine. If you are still not able to figure out let me know and I will try my best to help you out.

Best,
Meet Bhagdev
Program Manager, Microsoft

dharmdip:
Hi @meet-bhagdev, 

Thanks for your response, i am trying with FreeTds. I am checking on Windows server 2012 R2. I keep getting error ` Adaptive Server is unavailable or does not exist (TinyTds::Error) `, i went through firewall but it seems fine also i am able to connect from SQL management studio. 

Any thought?

Thanks, 
Dharmdip

meet-bhagdev:
Hi @dharmdip are you trying to connect from tiny tds or FreeTDS natively?

dharmdip:
Hello Meet,

Yes, I am trying to get this work with FreeTds natively.

dharmdip:
Hi @metaskills, 

Do you have any idea for this?

metaskills:
No... but please open a new issue and start there. This talk should not be in this issue.

dharmdip:
Yes, sure. 

Thank you !

PR #208, opened 2015-07-16
Ensure zero terminated strings, where C-str pointers are expected.

larskanis:
Starting with Ruby-2.2 strings can not be expected to be zero terminated.
It is therefore required to retrive String objets per StringValueCStr(). See https://github.com/ruby/ruby/blob/v2_2_0/NEWS#L319

Using StringValuePtr() without RSTRING_LEN() can lead to weird errors like:
http://stackoverflow.com/questions/28891447/weird-bytes-added-to-attribute-after-save-in-rails


metaskills:
Merged and made a change log entry. Thanks so much!

PR #213, opened 2015-08-01
Fix compat with Ruby <2.0.

larskanis:
Ruby-1.9 did use a mix of DL and Fiddle. That's why commit 3c90660598 broke compat with Ruby-1.9, when running on a Unix system with builtin FreeTDS.


metaskills:
Merged!

PR #217, opened 2015-08-09
[WIP] Fix Compile Warnings

metaskills:
I need some help with this please. I want to fix all the local and appveyor compiler warnings. However, my second commit 97353260 to fix the `implicit declaration of function 'rb_tinytds_raise_error'` opened up some new warnings WRT to `NOGVL_DBCALL` function parameters and return type (I think). 

This causes segmentation faults now perhaps related to string copies in that fix? Can someone take a look over that commit and give me some feedback?


larskanis:
You need to allocate some memory for error and source, if the client struct
stores pointers only. This could be done in the msg_handler, because
xmalloc is safe, even without GVL. And free'd after rb_str_new, before the
corresponding rb_raise().

#include "error.h" should be used for local header files.

I don't have a computer for the next week, beside a small Android device.
So this is what I read from the code, without being able to test anything.

metaskills:
Thanks @sgrif for doing the work in that other branch.

PR #224, opened 2015-10-10
Enable usage of OpenSSL for FreeTDS build.

### LINKED ISSUE #223: Ubuntu 10.10 uses FreeTDS 0.82 which is incompatible w/Azure
I know it's pretty ancient at this point, but I have a box running Ubuntu 10.10 and when I do

sudo apt-get --assume-yes install freetds-dev freetds-bin

it installs FreeTDS 0.82, which doesn't have support for Azure connections.  Is there any way I can force it to use 9.1, or installed 9.1 some other way?

TIA

larskanis:
This should fix #223 . The build will break, if no OpenSSL header files are installed.

These options for TLS are available :

1. Use OpenSSL for TLS:
   apt-get install libssl-dev
   gem inst tiny_tds

2. Use GnuTLS for TLS:
   apt-get install libgnutls-dev
   gem inst tiny_tds -- --enable-gnutls

3. Don't use TLS:
   gem inst tiny_tds -- --disable-openssl

Windows binary gems are unchanged - OpenSSL is always used.

### DIFF
--- a/ext/tiny_tds/extconf.rb
+++ b/ext/tiny_tds/extconf.rb
@@ -44,6 +44,12 @@ usage: ruby #{$0} [options]
     --enable-lookup
       Search for freetds through all paths in the PATH environment variable.
 
+    --disable-openssl
+      Disable OpenSSL for freetds build. No effect on system-freetds.
+
+    --enable-gnutls
+      Use GnuTLS instead of OpenSSL for freetds build.
+
     --enable-cross-build
       Do cross-build.
 HELP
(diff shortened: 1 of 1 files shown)


metaskills:
@larskanis Can you take another look at this. I could not get the gem to compile with all deps on my platform (Mac). I am using `rake -- --disable-system-freetds --disable-system-iconv --disable-system-openssl` for my command. I also noticed that CI is now breaking too for windows. Do you think platforms like Mac/Linux/Etc should use the GnuTLS option as the default?

larskanis:
I'll have a look at it, today.

metaskills:
Thanks @larskanis, I am at Microsoft this week and we are working on this today... 

https://twitter.com/metaskills/status/654692970344542209

metaskills:
And by that I mean waiting for you to take a look :)

larskanis:
> I am using rake -- --disable-system-freetds --disable-system-iconv --disable-system-openssl for my command.

Which error do you get? Why don't you want to use system-iconv?

> I also noticed that CI is now breaking too for windows.

[This](https://ci.appveyor.com/project/rails-sqlserver/tiny-tds/build/0.7.0.75) looks good so far. It's annoying, that the tests take ages to run... So please don't forget to request a Linux version of the SQL Server at Microsoft!

> Do you think platforms like Mac/Linux/Etc should use the GnuTLS option as the default?

Not sure. I took OpenSSL as default, because it is more or less a core library on Linux, so that it is easy to install on any distribution. GnuTLS has two dependencies (libgmp and libnettle) and is less standard - that makes it potentially more difficult to install.

How is the situation on OS X?

metaskills:
> Which error do you get? Why don't you want to use system-iconv?

I want to test a full local. The idea is that if needed, we can build a full static lib for any platform.

> It's annoying, that the tests take ages to run... 

Yea... that is the compile time mostly. We also upgraded our AppVeyor account today to help give us more parallel builds. So that should help a little... but yea, the benefit of the slowness is we get to test the full compile on that platform.

> please don't forget to request a Linux version of the SQL Server at Microsoft!

DONE! All the way to the top too :)

> How is the situation on OS X?

Still not able to compile in both this branch and the 0.95.20 work you did. See below, always fails with OpenSSL. Thoughts?

```
$ bundle exec rake -- --disable-system-freetds

checking compile-time options
configure: odbc disabled
checking if openssl is wanted... yes
configure: error: Cannot find OpenSSL libraries
```

```
$ bundle exec rake -- --disable-system-freetds --enable-system-openssl

checking compile-time options
configure: odbc disabled
checking if openssl is wanted... yes
configure: error: Cannot find OpenSSL libraries
```

metaskills:
My git-fu must have gone wrong and somehow I pulled your commit into master. Thanks Lars! Closing this issue.

PR #228, opened 2015-10-14
Allow CI to use SQL Azure. 

metaskills:
Takes advantage of Appyveor encrypted ENV vars. Even obfuscated the host just in case. We may have to get @larskanis PR of #224 in first for this to work.


metaskills:
I just did a rebase and force push. That means AppVeyor should start another build

PR #231, opened 2015-10-14
Update tiny_tds_ext.h

ajlam:
This is @ajlam fixing @sgrif breaking the build :arrow_down_small: :microphone:

### DIFF
--- a/ext/tiny_tds/tiny_tds_ext.h
+++ b/ext/tiny_tds/tiny_tds_ext.h
@@ -6,7 +6,10 @@
 
 #include <ruby.h>
 #include <ruby/encoding.h>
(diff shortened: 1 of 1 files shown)


sgrif:
Y'all are jerks and shouldn't support 1.9.3 which is EoL

metaskills:
Now thats a good point... I remember getting a lot of grief when I dropped 1.8 support. Not sure it is time for this one yet... : /

metaskills:
Damn.. what bothers me is how long AppVeyor is taking to START running the tests. The last commit to master was 16 hours ago... the build for that started ~4 hours past the commit. We are paying for a good queue level too I think, so I will double check into that.

PR #232, opened 2015-10-15
Update FreeTDS to 0.95.20 and use bz2 download

larskanis:
Not sure, whether there is a reason to use 0.91.112 ?

Unfortunately there is an issue with 0.95.20: It makes use of `inet_pton()`, but this is not yet supported in the old RubyInstaller's DevKit version 4.7.2. In order to work around this issue, I added a patch to freetds, that implements it as a wrapper to the `ws2_32.dll`.

I currently have no Windows box to hand to test this - so this is WIP! I'll try it out in the next days.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -91,7 +91,7 @@ end
 
 # Bundle the freetds sources to avoid download while gem install
 task gem_build_path(gemspec) do
-  add_file_to_gem(gemspec, "ports/archives/freetds-0.91.112.tar.gz")
+  add_file_to_gem(gemspec, "ports/archives/freetds-0.95.64.tar.bz2")
 end
 
 desc "Build the windows binary gems per rake-compiler-dock"
--- a/Rakefile
+++ b/Rakefile
@@ -99,6 +99,6 @@ task 'gem:windows' do
   require 'rake_compiler_dock'
   RakeCompilerDock.sh <<-EOT
 #    bundle install &&
-    rake cross native gem RUBY_CC_VERSION=1.9.3:2.0.0:2.1.6:2.2.2
+    rake cross native gem RUBY_CC_VERSION=2.0.0:2.1.6:2.2.2
   EOT
 end
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ The API is simple and consists of these classes:
 
 ## Install
 
-Installing with rubygems should just work. TinyTDS is currently tested on Ruby version 1.9.3 and upward.
+Installing with rubygems should just work. TinyTDS is currently tested on Ruby version 2.0.0 and upward.
 
 ```
 $ gem install tiny_tds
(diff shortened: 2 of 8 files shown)


metaskills:
Whoa! OK... a few things. I thought FreeTDS 0.91 was the stable... but it appears that I am behind the times. Seems 0.95 is now stable. I just checked the download on their homepage. That means we should drop 0.89 from the tests and start testing 0.91 and 0.95 now as the only two? FYI, I never had luck with 0.92, it always errored out.

I am testing this branch now...

metaskills:
So I've noticed that 1.9.3 seems to be an issue on AppVeyor with this work and that removing it got the build further. If we want, we can remote 1.9.3 from the appveyor.yml and start to remove 1.9.3 in general.

metaskills:
I've also noticed that an encoding test is failing now, that maybe what you are working on too. 

I am still trying to do some local builds using MacPorts for the `0.95.20` version. Hard to do on a plane tho :/ Also... I noticed that there is a `0.95.64` that I might test too. Did we want to explore that one?

larskanis:
I'm trying to get 0.95 running on Windows. That's why I triggered appveyor. But I don't work on encoding things, right now. 0.95.64 is brand new, so I didn't noticed it. I'll update the PR accordingly.

And on OSX with MacPorts it still doesn't build? Did you install `port install openssl` or `gnutls` and both are not found?

metaskills:
The MacPorts issue is my own and is likely due to :airplane: internet. I went that direction because I could not compile locally in this branch.

larskanis:
I'm more or less finished with the build part of this issue. The [appveyor test result](https://ci.appveyor.com/project/larskanis/tiny-tds/build/0.7.0.35/job/st5k3letsgbr2abd) isn't clean, but I'm less experienced with MSSQL, so it's maybe better if you (@metaskills) could have a look at it.

I noticed two small issues with freetds, to which I opened a PR for, [here](https://github.com/FreeTDS/freetds/pull/25) and [here](https://github.com/FreeTDS/freetds/pull/26) . Another issue is [this](https://github.com/larskanis/tiny_tds/commit/2d8b35568fff9ef27830aa1efede9ac229020ca7) , which is currently patched, but I think it is too special to the RubyInstaller environment (DevKit-4.7.2), to be worth a PR in FreeTDS.

Rakefile
```diff
@@ -91,7 +91,7 @@ end
 
 # Bundle the freetds sources to avoid download while gem install
 task gem_build_path(gemspec) do
-  add_file_to_gem(gemspec, "ports/archives/freetds-0.91.112.tar.gz")
+  add_file_to_gem(gemspec, "ports/archives/freetds-0.95.64.tar.bz2")
 end
```
metaskills: Very interesting... was this done to conform to the license?

metaskills: NM, saw the comment! :)

larskanis: Sorry, I don't understand your question. This is to include the standard freetds tar file into the gem, so that a simple `gem install` without any further settings can build freetds without an additional download. Any other version will be downloaded automatically, but this particular version is bundled and the download is skipped.

metaskills:
Closing this pull request in favor of #233.

PR #233, opened 2015-10-18
Default FreeTDS to 0.95. Support 0.91 Alternate.

metaskills:
This work was started by Lars Kanis <lars@greiz-reinsdorf.de>
Comments below are his from initial #232 pull request.

> I'm more or less finished with the build part of this issue. The AppVeyor results are not clean, but I'm less experienced with MSSQL, so it's maybe better if you (@metaskills) could have a look at it. I noticed two small issues with FreeTDS, to which I opened a PR for:
> 
> * https://github.com/FreeTDS/freetds/pull/25
> * https://github.com/FreeTDS/freetds/pull/26
> 
> The `0001-mingw_missing_inet_pton.diff` in the initial work does not have a FreeTDS pull request. I think it is too special to the RubyInstaller environment (DevKit-4.7.2) and no worth a PR in FreeTDS.

PR #240, opened 2015-11-19
Rename inet_pton patch, so that it's applied to freetds-0.95.67.

larskanis:
Regarding https://github.com/rails-sqlserver/tiny_tds/pull/233#issuecomment-158030934


metaskills:
Cool, can this be committed directly to the `freetds_095` branch?

larskanis:
I just pressed the "merge" above...

Issue #238, opened 2015-11-16
Reduce Appveyor Warnings


metaskills:
I'd like to see the Appveyor builds with less of these errors when compiling TinyTDS' extension.

```
warning: ISO C90 forbids mixed declarations and code [-Wdeclaration-after-statement]
```

I read that we can bypass this by using `-std=c99`. @larskanis do you know where I can put that? I tried putting things in the extension tasks' `config_options`, but that did not work.

larskanis:
@metaskills Shall I fix the warnings (add forward declarations for variables) or do you want to switch to c99? IMHO I would not change to c99, to be more portable.

metaskills:
Based on the c99 recommendation, I think we could just ignore this.

metaskills:
This is now done. Thanks so much @larskanis

PR #241, opened 2015-11-19
Fix warnings (PR#238)

larskanis:
Related to https://github.com/rails-sqlserver/tiny_tds/issues/238


larskanis:
The only warning that is left is about an unused variable [here](https://github.com/rails-sqlserver/tiny_tds/blob/master/ext/tiny_tds/result.c#L453). I'm not sure, if this return code can simply be ignored. Although in the [next function](https://github.com/rails-sqlserver/tiny_tds/blob/master/ext/tiny_tds/result.c#L464) it is ignored as well...

metaskills:
THANK YOU SO MUCH! This would have taken me forever.

Issue #227, opened 2015-10-10
Leverage FreeTDS Binaries In Native Gems


metaskills:
FreeTDS includes the following binaries and a [description of each](http://www.freetds.org/userguide/usefreetds.htm).

* bsqldb
* data copy
* defncopy
* fiscal
* freebcp
* osql
* tdspool
* tsql

We area already using `defncopy` in the SQL Server adapter for the schema dumper, but it relies on that bin being in the $PATH. These binaries are not findable for Windows builds. By including these in the TinyTDS gem, we can leverage scoped bin wrappers specific to the FreeTDS version built too.

**Do we wrap bins using a TinyTDS scoped prefix to denote these are our bins for just mirror? For example `defncopy` or `tinytds_defncopy`?**

Here are some links I gathered on this topic. 

* https://github.com/sstephenson/rbenv/wiki/Understanding-binstubs
* https://github.com/jekyll/jekyll/blob/master/jekyll.gemspec
* https://github.com/sj26/mailcatcher/blob/master/mailcatcher.gemspec
* http://guides.rubygems.org/specification-reference/

metaskills:
Closing for pull request #251. Only focusing on tsql now.

PR #251, opened 2015-12-13
Local ruby binstubs `./bin/tsql` for `./exe/tsql` executable.

metaskills:
See #227 for full details.

* Copy `tsql` exe during FreeTDS compile install.
* Use ptools gem dep for easy cross platform `File.which` support.
* Local binstub will prefer local gem exe but delegate to system as needed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,5 +15,6 @@ lib/tiny_tds/tiny_tds.rb
 Gemfile.lock
 misc
 *.gem
+/exe/tsql
 /ports/*
 !/ports/patches/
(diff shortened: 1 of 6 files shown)


larskanis:
@metaskills Good idea and well done!

metaskills:
Thanks! I figured if we did more of these, then additions would be easy. We could even extract much of the bulk form bin/tsql to some sort of lib/tiny_tds/bin_helper.rb if needed.

metaskills:
Finally got it working with the gem:windows task now too https://github.com/rails-sqlserver/tiny_tds/commit/351e7d8cbe514144a6e85957c72f6956230d90f1

Issue #252, opened 2015-12-16
Fix Rake Gem Windows


metaskills:
@larskanis Could you please take a look at the `rake gem:windows` task? I get the following error when compiling. https://gist.github.com/metaskills/de0beec0c6a990a0870d

PR #253, opened 2015-12-16
Update to mini_portile2 and fix packaging of freetds tar as bz2.

larskanis:
This should fix https://github.com/rails-sqlserver/tiny_tds/issues/252 . 

I was able to reproduce the "undefined method `include?' for nil:NilClass" error in #252 . But after the update to mini_portile2 it magically disappeared and `rake gem:windows` works fine now.

As a side note: rake-compiler-dock-0.5.0 will support ruby-2.3 as well. I'll release it when ruby-2.3.0 is released, so that binary gems can be built.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -86,14 +86,14 @@ end
 
 # Bundle the freetds sources to avoid download while gem install.
 task gem_build_path(gemspec) do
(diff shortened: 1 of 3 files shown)


metaskills:
THANK YOU! New beta gems published. Now on to some adapter work.

PR #257, opened 2016-01-15
Add patch for getopt

l8nite:
I was trying to get the latest to install on alpine linux, stole this from http://git.alpinelinux.org/cgit/aports/tree/main/freetds/fix-includes.patch

### DIFF
--- a/ports/patches/freetds/0.95.78/0002-missing-getopt-include.diff
+++ b/ports/patches/freetds/0.95.78/0002-missing-getopt-include.diff
@@ -1,0 +1,12 @@
+diff --git a/src/apps/tsql.c b/src/apps/tsql.c
+index 9015262..3ff9475 100644
+--- a/src/apps/tsql.c
++++ b/src/apps/tsql.c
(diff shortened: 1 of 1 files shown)


larskanis:
Wouldn't it be better to open a pull request on freetds? The patches in tiny_tds are usually only applied for issues, that are very specific to the environment in which tiny_tds is typically used (like the old gcc/mingw version of the RubyInstaller's Devkit).

l8nite:
@freddy77 what do you think?

metaskills:
Correct, the patches are there to primarily support our efforts to precompile static libs. They are only ever used when FreeTDS is not installed on your system via some package manager. For example, I use MacPorts and there may be a few patches we do for our platform, TinyTDS should not assume that, I think your package manager and/or relationship to the parent project, in this case FreeTDS should. Could you open an issue on http://github.com/FreeTDS/freetds and link it to here?

l8nite:
Thanks guys, I've opened https://github.com/FreeTDS/freetds/issues/45

PR #262, opened 2016-02-08
Update README.md

aharpervc:
Each of these features is stated independently, but they work well together. I actually forgot this was possible until recently.


metaskills:
Love it!

PR #270, opened 2016-04-29
Add ruby 2.3 to appveyor config

aharpervc:
expecting this to work per https://www.appveyor.com/docs/installed-software#ruby

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -46,6 +46,8 @@ environment:
     - ruby_version: "200"
     - ruby_version: "22"
     - ruby_version: "22-x64"
+    - ruby_version: "23"
+    - ruby_version: "23-x64"
 on_failure:
   - find -name compile.log | xargs cat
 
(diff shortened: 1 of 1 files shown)


aharpervc:
Is there an intentional reason ruby 2.1 isn't on the list?

metaskills:
Testing all the rubies made for super slow tests and feedback. I picked what felt reasonable.

metaskills:
My thinking was Ruby 2.0 was the bottom... and that 2.1 was not that different between 2.0 or 2.2 and one or the other would catch 2.1 in some way. Hope that makes sense.

aharpervc:
I see. I don't know much about appveyor... maybe there's a way to do a "lightweight" test on just 1 ruby (latest version), then separately/manually a "full" test on all supported rubies?

metaskills:
Well, what we are testing is talking to different SQL Server versions. The matrix of all ruby versions to all SQL Server versions has diminishing returns. I'd like to keep the test on the bare minimum of Ruby versions to get that done. To that end, can this PR remove `ruby_version: "22"`?

aharpervc:
I get it, so the concept is to just put the min/max versions, not all intermediate versions

metaskills:
Merging... seems the branch merge tests have not been passing on Appveyor lately for some reason. Will watch master build now and look at the other PR.

aharpervc:
Sweet

PR #292, opened 2016-07-18
Always set database name in login packet

anderscarling:
Allows support for contained users in recent SQL Server versions.

Running the specs from OS X against SQL Server 2016 (on remote windows server obviously)
renders one test failure, however the same failure is also present on `master`:
```
Basic query and result#test_0015_must insert and find unicode data [/Users/lowe/src/tiny_tds/test/result_test.rb:131]:
```

It's probably important to run on appveyor etc to make sure we don't break compatibility with old SQL Server versions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,7 @@ Creating a new client takes a hash of options. For valid iconv encoding options,
 * :timeout - Seconds to wait for a response to a SQL command. Default 5 seconds.
 * :encoding - Any valid iconv value like CP1251 or ISO-8859-1. Default UTF-8.
 * :azure - Pass true to signal that you are connecting to azure.
+* :contained - Pass true to signal that you are connecting with a contained database user.
 
 Use the `#active?` method to determine if a connection is good. The implementation of this method may change but it should always guarantee that a connection is good. Current it checks for either a closed or dead connection.
 
--- a/ext/tiny_tds/client.c
+++ b/ext/tiny_tds/client.c
@@ -3,7 +3,7 @@
 
 VALUE cTinyTdsClient;
 extern VALUE mTinyTds, cTinyTdsError;
-static ID sym_username, sym_password, sym_dataserver, sym_database, sym_appname, sym_tds_version, sym_login_timeout, sym_timeout, sym_encoding, sym_azure;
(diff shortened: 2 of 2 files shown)


anderscarling:
We should probably add some tests to this as well, but I'm not sure how you would like to go about it as contained users are only supporter in sql server 2012 and later.

anderscarling:
If I read the app veyor results correctly it seems this breaks down completely on sql server 2008.

@metaskills How would you feel about adding a configurable option (something like `user_is_contained` or `contained_database` or something like it that triggers the behavior)?

metaskills:
> How would you feel about adding a configurable option

Love it. Since every option when connecting is about the database and we carry that implied subject. May I suggest just using `contained` boolean for the flag?

anderscarling:
Cool, I'll go ahead and try to push something like that in the next few hours.

Is there any reasonable way to write a test for that behavior? I don't think it's realistic to add a contained user to the default test schema, as it (might) require config changes on the server running the test etc.

metaskills:
Agreed... the test would not feel super portable.

anderscarling:
Pushed a new version of the branch.

It doesn't break anything new when running tests locally, and it shouldn't change any behavior unless `contained: true` is passed with the options.

@metaskills Would you like anything added for you to be able to merge it?

metaskills:
> Would you like anything added for you to be able to merge it?

Not off hand. Will look over it a bit later.

anderscarling:
I'm working on the project related to this change today, so if you have any feedback on the PR i would appreciate it or i will probably not have time to look at it until next week.

anderscarling:
@metaskills - was there any issues with this? I would really love to get it merged! :)

metaskills:
Can we rebase to master and see if tests still pass?

anderscarling:
I'm not setup to run the tests locally atm, but i pushed the rebased branch.

metaskills:
> I'm not setup to run the tests locally atm, but i pushed the rebased branch.

That's cool! I got Appveyor setup for us so we can just watch the build here. Takes a bit tho.

anderscarling:
Looks likes the test are still green!

metaskills:
I do love this, maybe one day we will deprecate `:azure` and just use the new `:contained`. Merging now. Thanks so much for your work and sorry for slowness. I'll work at getting v1.0.5 out soon.

PR #293, opened 2016-07-18
Error messages greater than 1024 chars generates a buffer overflow

luisrpp:
We have some stored procedures implemented in C#, deployed as a DLL on SQL Server.

In some scenarios, an exception is thrown and the error message + the stack trace of the error generates a message greater than 2048 characters.

As the fields `error` and `source`, of the `tinytds_errordata` struct, are defined with size 1024, it generates a buffer overflow error for messages greater than this value, which ends up killing the process.

In this PR, the suggestion is to use `strncpy` instead of `strcpy`, to shrink the message if it is greater than a defined size.

[logs.zip](https://github.com/rails-sqlserver/tiny_tds/files/370295/logs.zip)

### DIFF
--- a/ext/tiny_tds/client.c
+++ b/ext/tiny_tds/client.c
@@ -110,8 +110,8 @@ int tinytds_err_handler(DBPROCESS *dbproc, int severity, int dberr, int oserr, c
     */
     if (!userdata->nonblocking_error.is_set) {
       userdata->nonblocking_error.cancel = cancel;
-      strcpy(userdata->nonblocking_error.error, dberrstr);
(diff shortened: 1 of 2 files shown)


metaskills:
Thanks for doing this work! I'll find some time soon to help test this and get it merged in.

luisrpp:
Cool! Just to let you know, I was only able to reproduce this error in a Ubuntu box.

metaskills:
I merged this in via another branch. Thanks! Will be in the upcoming 1.0.5 release.

Issue #288, opened 2016-06-23
 Connect Sql server in Ruby on rails #489 


luis77:
I need to connect to a SQL Server database, I installed gem 'tiny_tds' gem 'activerecord-sqlserver-adapter' gem 'deprecated'
my yml file:

default: &default
adapter: sqlserver
host: basddatos.dyndns.org
username: zu
password: pass
pool: 5
timeout: 5000

development:
<<: *default
database: Basdatos_2016

I get this error: TinyTds::Error Server name not found in configuration files WHY? thanks

metaskills:
I see this is moved from https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/issues/489#issuecomment-228059953

Have you configured SQL Server to allow for TCP connections. I suspect his has nothing to do with TinyTDS and you are actually debugging if you can reach your box with DNS or some SQL Server settings.

luis77:
let me check, but Do I need more code than I have?

metaskills:
No, that should work just fine if you can reach the host and have SQL Server set to listen on via TCP on the host/port.

luis77:
Apparently everything is fine. I don't know where the problem is, sql server is set to listen on via TCP but still the same error 
TinyTds::Error
Server name not found in configuration files

metaskills:
Did you try to remove Dydns.org from the equation?

Karamorf:
I think I ran into this same problem, however without the rails part. It seems to not work in tiny_tds versions 0.9.5.rc.3 and 1.0.4 however it did work in 0.7.0. This was the ruby I was running:
ruby -e "require 'sequel'; db = Sequel.connect({host: '10.198.135.174', database: 'database',user: 'user',password: 'PASSWORD',adapter: 'tinytds',port: 1433, azure: true}); puts db[:some view].where(:uid => 'some_username').all"

This is the output I'd get for the error:
/home/user/.gem/ruby/2.3.0/gems/tiny_tds-0.9.5.rc.3/lib/tiny_tds/client.rb:43:in `connect': TinyTds::Error: Login failed for user 'idr_admin@10'. (Sequel::DatabaseConnectionError)

Or with the domain name (when host was set to 'name.domain.com'):
[...]

PR #299, opened 2016-09-06
Adds a simple_username option for case where azure is true but you don't want user@domain

austra:


austra:
This was an issue with an out of date freetds, appears to be resolved after updating from freetds 0.95 to 1.0

metaskills:
Thanks for the PR and followup!

austra:
We ended up needing this after all.  It looks like it might be the same issue as #288  (or at least the[ hijacked issue](https://github.com/rails-sqlserver/tiny_tds/issues/288#issuecomment-232682509) )

metaskills:
@austra Can you elaborate a bit more, linked articles, etc on how and when using Azure does not enforce that username format?

anderscarling:
If I remembwr correctly this would be functionally equivalent of running the new `:contained` option added in #292

austra:
Just getting back around on this, thanks @anderscarling , you are correct.

PR #303, opened 2016-09-27
Ensures numeric options are treated numerically

kenaniah:
Allows for numeric options in URI connection strings. 

E.g. `tinytds://user:passwd@host:1433/database?timeout=60`


metaskills:
I just did some changes to master, can you rebase your branch so we can see how the tests run?

kenaniah:
Rebased. Looks like the tests now pass.

metaskills:
I merged this in and did a small change to ensure we do not cast nil to 0. Hope this makes sense and still works for you. https://github.com/rails-sqlserver/tiny_tds/commit/3f445226301baeaf8eb695937702b457f9594eab

PR #314, opened 2016-10-22
Don't reset use_utf16 option, when initializing with false

vijay-koppala:
||= is a deceiving operator for Boolean values. It could mean, assign-if-nil or assign-if-false. It is just a short-hand for `a = a || new_value`


metaskills:
Thanks... maybe we should use idiomatic Ruby instead. Would this work?

```ruby
opts[:use_utf16] = opts.fetch(:use_utf16) { true }
```

metaskills:
Actually, I like this better. Merged in :)

PR #320, opened 2016-11-27
Update dependences. 

metaskills:
Use FreeTDS 1.00.21.


coderjoe:
@metaskills is there anything you want to keep in this PR?

metaskills:
Yes, I'd like everything here except for the parts you changed. I can try squashing the commits here and pushing up just the new bits. The idea would be that we want to test Ubuntu with the versions of OpenSSL and FreeTDS just like Appveyor.

PR #321, opened 2016-11-27
Add DB col type `98` for int for `LOGINPROPERTY` function.

metaskills:


metaskills:
I am going to move our `SYBVARIANT` case to the bottom of `rb_tinytds_result_fetch_row` and check the length of the data. If == 4 then we will assume a DBINT. If not, we will pass it thru as a binary string.

Issue #310, opened 2016-10-09
Slow Connections; Not valid win32 Application; OpenSSL


metaskills:
We have had several issues with OpenSSL statically linked in the Windows gem. Issues include #250 and #290. Symptoms of using OpenSSL v1.0.x include:

* Slow initial connections.
* tiny_tds.so not a valid win32 application

### Potential Workarounds

* Adding TinyTDS just before Rails starts up, for example above Bundler's setup in boot.rb.
* Setting chcp 866 for tiny_tds.so

### Proposed Next Steps

One of the following two have been proposed.

* Hacks for old OpenSSL. Like using libeay32.dll.
* Use a modern OpenSSL v1.1.0

##### Test Appveyor Build

It would be really nice if we changed our OpenSSL version in extconsts.rb to use 1.1.0.x. I found when doing this that there are some Perl errors on Appveyor because the version is too low. We have a path config set to `SET PATH=C:\MinGW\msys\1.0\bin;%PATH%` and that seems to put a old version of Perl in. I did try adding `SET PATH=C:\Perl\bin` after the MinGW, but it does not seem to add totally fix the issue.

##### Update Docker Usage

I will be doing some CircleCI testing to ensure 1.1.x works for posix systems. However, there is a fair amount of work to make the native `rake gem:windows` tasks work for us to distribute native Windows gems.

YvesR:
Hello @metaskills,
did your tests against Azure pass lately?
I tried 1.0.5 and AR 4.2 and still get the same problem. First connect fails, seconds works then. Also to mention that the 1.0.2g open-ssl still has this CPU bug that loading rails console takes 5 times compared to the fixed libeay32.dll.

bye Yves

nmscholl:
@metaskills 
I am currently experiencing the TinyTDS  Not valid win32 Application error when executing a chef run. In some contexts, I have been able to resolve the problem by requiring TinyTDS in a particular order with the other requirements that use the conflicting components, however I have encountered a situation where this is not sufficient, I still get the error. I attempted to use chcp to resolve the issue as you suggested, however I am unsure what is meant by setting it up "for" a file. Is that different than running it in the same powershell window where tinyTDS will be used? In any case, chcp 866 was used in said powershell window, chef was executed, and the error still occurred. Any information you can provide would be appreciated.

coderjoe:
[...]

PR #322, opened 2016-11-30
openssl 1.0.2j, freetds 1.00.21, and openssl shared lib generation

coderjoe:
So with respect to issue #310  I was running into 3 distinct problems

1. When running tiny_tds in half of our 64 bit and 32bit windows environments tinytds could not load due to being an invalid win32 executable. (The provided workarounds did not work for us)
2. After updating to openssl 1.0.2j, on 64 bit windows environments I received an error: `998: invalid memory access` in windows
3. Builds would intermittently fail on appveyor for ruby 2.2x64

All three of these problems seemed to boil down to the `dllwrap` dll creation in the native build process. In my searches I found numerous references to `dllwrap` occasionally causing problems when dll relocation occurs.

To test I removed the custom DLL generation, removed the symbol stripping (for debugging purposes) and rebuilt. Upon rebuild all tests pass, and when used on an instance experiencing `invalid win32 application` it proceeds as normal.

Now I'm not entirely sure why tinytds is building its own openssl shared libraries but letting openssl generated the libraries seems to have solved my problems. Thoughts?


Rakefile
```diff
@@ -68,20 +68,20 @@ Rake::ExtensionTask.new('tiny_tds', gemspec) do |ext|
...
     ]
     # We don't need the sources in a fat binary gem
     spec.files = spec.files.reject { |f| f =~ %r{^ports\/archives/} }
     spec.files += dlls.map { |dll| "ports/#{host}/bin/#{File.basename(dll)}" }
     spec.files += Dir.glob('exe/*')
-    dlls.each do |dll|
-      file "ports/#{host}/bin/#{dll}" do |t|
-        sh 'x86_64-w64-mingw32-strip', t.name
-      end
-    end
```
coderjoe: I removed this stuff so I could use the symbols while debugging and it _seemed_ like some symbols which looked useful for relocation may also have been removed... but I couldn't confirm so I left it removed...

ext/tiny_tds/extconf.rb
```diff
@@ -157,7 +157,7 @@ def configure
         end
         args = [ "CFLAGS=-DDSO_WIN32",
             "./Configure",
-            "no-shared",
+            "shared",
```
coderjoe: I'm really not sure about the implications of this change. The resulting binaries have different exports, different sizes, and different symbols, and use a different build process than the old `dllwrap`. Given the comment below I'm not sure how safe this is. Any ideas?

metaskills:
This is amazing work. I honestly need to rely on you to make the calls on this because this area of TinyTDS is most unfamiliar to me. 

Is there anything else you want to do before I merge this? Are there any next steps after this? I am hitting Rails v5 in the adapter hard, could you help with moving the adapter to the new miniportile docker work so we can build and push new Windows gems?

metaskills:
@coderjoe I added you as a collaborator. Feel free to merge when ready. Super appreciate the help!

coderjoe:
Thank you very much for your incredible responsiveness in helping me resolve this issue especially around the holidays. I'm glad it has been helpful. 

I've got some changes I want to make before committing, so I'm going to close this PR and continue to work my forked branch so I can make changes without fear of them accidentally being merged. I'll submit a new PR once it's cleaned up, re-tested, and ready to go. :)

I don't have much availability tomorrow to work on opensource stuff, but I hope to be able to put get everything together ASAP probably this weekend at the latest.

metaskills:
Either works fine, you could have kept this open and made commits to see the builds run. Leaving it open would also mean others could find it more easily and comment. Don't sweat the churn to this branch either. Once you are ready, we can do a squash merge to keep it clean. Either way, whatever works best for you.

coderjoe:
Oh that's a really good point. I'll re-open for visibility's sake. Thanks for the advice. 😄

coderjoe:
@metaskills do you know of any way to convince appveyor to rebuild a commit without re-committing? It failed on my recent commits but it looks like it might be a network error and nothing more.

metaskills:
> do you know of any way to convince appveyor to rebuild a commit without re-committing? 

Sadly not without logging into Appveyor and clicking re-build. But logging in is limited to the #rails-sqlserver account where I set that up. So no-op commits are best.

The state of this PR looks fine. Can I merge this work in?

coderjoe:
@metaskills presuming this last test run passes and you can't duplicate any failures on Azure (I haven't had time to test it yet myself) then I'm good to merge. :)

metaskills:
Hit the button whenever you are ready :) I recommend a squash merge with a wrapped up commit message. I'm starting work in the adapter now. Trying to fix the forgetting assignment bug.

Issue #333, opened 2017-01-10
gem v1.1.0 fails to install on openSUSE due to lib64 naming conventions


mbixby17451:
Installing the latest v1.1.0 gem on openSUSE (Leap 42.2) fails because although the bundled freetds correctly builds into a "lib64" subdirectory per openSUSE naming conventions, the gem's extconf.rb is hard-coded to look for freetds in a "lib" subdirectory.

Output from gem install:

```
markb@linux:~> gem install tiny_tds
Fetching: mini_portile2-2.1.0.gem (100%)
Successfully installed mini_portile2-2.1.0
Fetching: tiny_tds-1.1.0.gem (100%)
Building native extensions.  This could take a while...
ERROR:  Error installing tiny_tds:
        ERROR: Failed to build gem native extension.

    current directory: /home/markb/stack/lib64/ruby/gems/2.3.0/gems/tiny_tds-1.1.0/ext/tiny_tds
/home/markb/stack/bin/ruby -r ./siteconf20170109-3991-kqksju.rb extconf.rb
extconf.rb:14: warning: already initialized constant OpenSSL::SSL::VERIFY_PEER
checking for sybfront.h... no
Downloading freetds-1.00.21.tar.bz2 (100%) 
Extracting freetds-1.00.21.tar.bz2 into tmp/x86_64-pc-linux-gnu/ports/freetds/1.00.21... OK
Running git apply with /home/markb/stack/lib64/ruby/gems/2.3.0/gems/tiny_tds-1.1.0/ports/patches/freetds/1.00.21/0001-mingw_missing_inet_pton.diff... OK
Running 'configure' for freetds 1.00.21... OK
Running 'compile' for freetds 1.00.21... OK
Running 'install' for freetds 1.00.21... OK
Activating freetds 1.00.21 (from /home/markb/stack/lib64/ruby/gems/2.3.0/gems/tiny_tds-1.1.0/ports/x86_64-pc-linux-gnu)...
checking for sybfront.h... yes
checking for sybdb.h... yes
checking for tdsdbopen() in -lsybdb... no
-----
freetds is missing.
Do you have FreeTDS 0.95.80 or higher installed?
-----
*** extconf.rb failed ***
Could not create Makefile due to some reason, probably lack of necessary
libraries and/or headers.  Check the mkmf.log file for more details.  You may
need configuration options.

Provided configuration options:
        --with-opt-dir
        --with-opt-include
        --without-opt-include=${opt-dir}/include
[...]

PR #334, opened 2017-01-10
lib64 awareness for openSUSE and other distros

mbixby17451:
Fix for issue #333 

Libraries on openSUSE and other distros end up in a lib64 subdirectory instead of in lib. If lib64 exists, use that for libraries, otherwise fall back to lib, to successfully find the bundled freetds.


metaskills:
Closing, please see #342 for details.

Issue #340, opened 2017-02-15
gem install fails behind firewall


coldnebo:
In a docker context and from a machine behind a firewall that restricts traffic to a gem proxy server and doesn't allow any other outbound requests, I see the failure below in the install logs.

I *think* what is going on is that if tiny_tds doesn't find freetds dev headers (and of course the lib) installed, it will attempt to fetch FreeTDS source (or binary?) and build/install it in the gem subdir.  If so, then the reason I see an `ECONNRESET` is because the outbound request for FreeTDS source/binary is being blocked by the firewall.

What do you think?

If so, the only fix I could suggest would be trapping the lowlevel failure to fetch and raising a clearer error: "Couldn't retrieve FreeTDS source/binary.  You must install FreeTDS manually before continuing and use the --with-freetds-{dir,include,bin} options"  (whatever the syntax actually is).

Thanks!

```111634 [INFO] Gem::Ext::BuildError: ERROR: Failed to build gem native extension.

    /usr/local/bin/ruby extconf.rb
extconf.rb:14: warning: already initialized constant OpenSSL::SSL::VERIFY_PEER
checking for sybfront.h... yes
checking for sybdb.h... yes
checking for tdsdbopen() in -lsybdb... yes
checking for dbanydatecrack() in -lsybdb... no
*** extconf.rb failed ***
Could not create Makefile due to some reason, probably lack of necessary
libraries and/or headers.  Check the mkmf.log file for more details.  You may
need configuration options.

Provided configuration options:
	--with-opt-dir
	--without-opt-dir
	--with-opt-include
	--without-opt-include=${opt-dir}/include
	--with-opt-lib
	--without-opt-lib=${opt-dir}/lib
	--with-make-prog
	--without-make-prog
	--srcdir=.
	--curdir
	--ruby=/usr/local/bin/ruby
	--help
	--enable-lookup
	--disable-lookup
	--with-freetds-dir
	--without-freetds-dir
	--with-freetds-include
	--without-freetds-include=${freetds-dir}/include
	--with-freetds-lib
	--without-freetds-lib=${freetds-dir}/lib
	--with-sybdblib
	--without-sybdblib
	--with-sybdblib
	--without-sybdblib
	--enable-system-freetds
	--disable-system-freetds
	--enable-system-iconv
	--disable-system-iconv
	--enable-system-openssl
	--disable-system-openssl
	--enable-gnutls
	--disable-gnutls
	--enable-openssl
	--disable-openssl
[...]

PR #341, opened 2017-02-16
Use alternate freetds source for 1.00.26

avit:
ftp://ftp.freetds.org has been down over a day.

This is a workaround to change the mirror (same one used by [homebrew's freetds formula][1]) for the current version of _freetds_ in order to fix broken builds.

I've also made the configuration a little more flexible to be able to change the source in the future if needed. (Maybe we could make this a list of mirrors to fall back on: I haven't investigated that yet.)

[1]: https://github.com/Homebrew/homebrew-core/blob/master/Formula/freetds.rb


avit:
Fixes #340

avit:
Hold back on this, the official source is back: I can revert the commit for changing the mirror for this version. Let me know if the other changes are worth keeping.

metaskills:
Closing, please see #342 for details.

PR #344, opened 2017-02-19
Use OpenSSL v1.1.0e & FreeTDS v1.00.27

metaskills:


metaskills:
Right now I am trying to figure out how to get the right Perl path in place. When we do `SET PATH=C:\MinGW\msys\1.0\bin;%PATH%` it puts an older version of Perl in the path. That can be fixed by doing `SET PATH=C:\Perl\bin;%PATH%` below it. And the `perl --version ` show the correct 5.22. However, MiniPortile still fails below because it finds that previous version. Any ideas on how to help? My Questions are:

* Just a gut check, but is adding `C:\MinGW\msys\1.0\bin` right? Should we do something else?
* Depending on that answer, how do we get `rake compile` to see the right version?
* Maybe this work should happen after #345 now?

YvesR:
You are trying this in a windows environment?
I am still working with docker image rake-compiler where freetds build failed as described. Do you fail earlier with your perl problem or later (means you can build freetds)?

metaskills:
> You are trying this in a windows environment?

I am trying to make all the CI builds first. Ubuntu seemed to work pretty easily. Focus is on Appveyor. That usually gives me a good signal to move onward to local rake compiler work.

orgads:
I'm also trying to build it (locally for now). Will post my own PR if I succeed.

metaskills:
Pull Request #347 pulled in OpenSSL v1.1.0 and we got new versions of TinyTDS at v1.2.0 which should solve this issue. Please reply back and let us know.

PR #345, opened 2017-02-20
Remove MiniPortile from the extension build

coderjoe:
As per #342 we don't want to automatically build and install FreeTds on all architectures. This should only be needed for the windows fat binary and possibly for local builds during development.

This PR does the following:

1. Adds a `rake ports` task which will build all the MiniPortile sources for use in a local build preventing the sorts of failures we saw in issue#340 and PR#341
2. Adds a `rake ports:cross` task which will cross-compile the ports for our windows target architectures
3. Removes all references of MiniPortile from the extension build removing hopefully preventing the class of errors we saw with issues like issue#333 and PR#334
4. It also switches over to using the standard OpenSSL shared library build instead of hacking together our own OpenSSL libraries.

The fourth point is worth some discussion. It definitely simplifies things on this end, and with ruby-installer and rails-installer getting closer to an up-to-date openssl build it doesn't seem that big a deal if the wrong one gets loaded. That said, once we go to OpenSSL 1.1.0 if it causes problems, we may need to revisit this portion of the change. I'm hoping will go smoothly.

PR #347, opened 2017-03-02
Openssl 1.1.0

orgads:
See discussion on https://github.com/rails-sqlserver/tiny_tds/issues/310 and previous PR https://github.com/rails-sqlserver/tiny_tds/pull/344 by @metaskills.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.diff eol=lf
--- a/Rakefile
+++ b/Rakefile
@@ -68,8 +68,8 @@ Rake::ExtensionTask.new('tiny_tds', gemspec) do |ext|
     gemplat = spec.platform.to_s
     host = platform_host_map[gemplat]
     dlls = [
-      "libeay32-#{OPENSSL_VERSION}-#{host}.dll",
-      "ssleay32-#{OPENSSL_VERSION}-#{host}.dll",
+      "libcrypto32-#{OPENSSL_VERSION}-#{host}.dll",
+      "libssl32-#{OPENSSL_VERSION}-#{host}.dll",
       'libiconv-2.dll',
       'libsybdb-5.dll'
     ]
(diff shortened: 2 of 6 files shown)


orgads:
Ok, this perl installation just doesn't work in this environment. Is there a way to grab another perl?

metaskills:
Love the DDL rename commit. 

> this perl installation just doesn't work in this environment

What does the OpenSSL team use on their Appveyor builds? This would help isolate if our `C:\MinGW\msys\1.0\bin` is still leaking issues.

orgads:
The problem is not AppVeyor, but the Ruby devkit. It is prepend to PATH, and users perl from mSYS1. This will be a problem with local builds as well.

I see 2 options:
# Download perl as part of openssl installing, and force using it (like I did here with C:\Perl, only that it will be in the build directory)
# Wait for rubyinstaller2, and drop support for devkit.

I prefer the first option.

The question is how do we fetch perl? Maybe from MSYS2? Do you have an idea?

metaskills:
I don't since this is not my speciality. cc @coderjoe

orgads:
Ok, this perl just doesn't work for openssl. I guess we have no choice but to install msys2.

orgads:
... Or use Git's! :)

orgads:
Finally a link problem! 64 but compiles. I'll look into it later.

orgads:
Woohoo! :)

metaskills:
Amazing Work. Thanks! I'd like to hold off on this till #345 is done. I see some git conflicts, but that can be worked out. 

In the meantime, I used this branch locally to build and publish `v1.2.0.beta1` and pushed those gems to rubygems.org. I'm going to solicit others to test it out. Again, thanks!!!

orgads:
My pleasure :)

metaskills:
@coderjoe What do you think of merging this in and coping with the rebase in #345? I think it would be cool to get a 1.2.0 version out that is confirmed working better on Windows.

coderjoe:
A hearty :+1:. The rebase shouldn't be too much trouble at all. It sounds worth it.

Issue #84, opened 2012-06-09
ActiveRecord Timeout Ignored When Using TinyTDS


jlaskowski:
I am running a Rake task under Rails 3.1.4, Ruby 1.9.3.  I am using tiny_tds (0.5.1 x86-mingw32).

Each time I run the rake task, I get a connection timeout on the same line of code.  Printing a timestamp to the console before the line executes and as soon as the rescue block is entered, I can see that 6 seconds have transpired.  Adding either of these lines to the database.yml didn't change it from 6 seconds:

timeout : 25
timeout: 25000

Also, when I execute the SQL using a different SQL Server client, it completes successfully in 5.4 seconds.

Here's the exception, https://gist.github.com/2908008

attilahorvath:
Sorry to resurrect this old thread, I was just curious if it's still impossible to set the timeout on a per-connection basis.

It seems this feature has landed in FreeTDS in the meantime (https://github.com/FreeTDS/freetds/commit/33fa016ca933ea5534007bce69c10e657e92e653) but according to my tests TinyTDS is not taking advantage of it yet.

Would it be possible to implement this feature now? I'm happy to help get it working.

attilahorvath:
It was actually easy to track this down, so I've submitted a PR for this: https://github.com/rails-sqlserver/tiny_tds/pull/348.

PR #348, opened 2017-03-07
Add support for separate timeout values per connection

attilahorvath:
As referenced in an old issue (https://github.com/rails-sqlserver/tiny_tds/issues/84), TinyTDS currently only supports a single, global timeout value because of a limitation in FreeTDS. However, support for per-connection timeout values has been added to FreeTDS in the meantime (https://github.com/FreeTDS/freetds/commit/33fa016ca933ea5534007bce69c10e657e92e653), so we can take advantage of it now.

For some reason the call to FreeTDS expects a string value instead of an integer, so I'm forcing the conversion. This way it remains backward compatible with existing code.

### DIFF
--- a/ext/tiny_tds/client.c
+++ b/ext/tiny_tds/client.c
@@ -326,8 +326,6 @@ static VALUE rb_tinytds_connect(VALUE self, VALUE opts) {
     dbsetlapp(cwrap->login, StringValueCStr(app));
   if (!NIL_P(ltimeout))
     dbsetlogintime(NUM2INT(ltimeout));
-  if (!NIL_P(timeout))
-    dbsettime(NUM2INT(timeout));
   if (!NIL_P(charset))
     DBSETLCHARSET(cwrap->login, StringValueCStr(charset));
   if (!NIL_P(database)) {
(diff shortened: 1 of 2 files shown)


lib/tiny_tds/client.rb
```diff
@@ -48,8 +48,10 @@ def initialize(opts = {})
       opts[:encoding] = opts[:encoding].nil? || opts[:encoding].casecmp('utf8').zero? ? 'UTF-8' : opts[:encoding].upcase
       opts[:port] ||= 1433
       opts[:dataserver] = "#{opts[:host]}:#{opts[:port]}" if opts[:dataserver].to_s.empty?
-      forced_integer_keys = [:login_timeout, :port, :timeout]
+      forced_integer_keys = [:login_timeout, :port]
       forced_integer_keys.each { |k| opts[k] = opts[k].to_i if opts[k] }
+      forced_string_keys = [:timeout]
+      forced_string_keys.each { |k| opts[k] = opts[k].to_s if opts[k] }
```
metaskills: Given that we likely continue to want this to be integers and passed down to `dbsettime(NUM2INT(timeout))` I would suggest removing this and doing the string coercion in client.c within the `dbsetopt` call.

ext/tiny_tds/client.c
```diff
@@ -326,8 +326,6 @@ static VALUE rb_tinytds_connect(VALUE self, VALUE opts) {
     dbsetlapp(cwrap->login, StringValueCStr(app));
   if (!NIL_P(ltimeout))
     dbsetlogintime(NUM2INT(ltimeout));
-  if (!NIL_P(timeout))
-    dbsettime(NUM2INT(timeout));
```
attilahorvath: I've fixed this behaviour in FreeTDS in the meantime: https://github.com/FreeTDS/freetds/pull/107. So from now it'll be possible to just simply call `dbsettime(timeout)` each time.

We should still use this original solution though to be compatible with existing versions.

metaskills:
@attilahorvath Sorry for the delay, I'd like to pull this in. In order to help me and other understand the final proposed changes – could you include a change to the README section `#tinytdsclient-usage` please?

attilahorvath:
Added a note to the `README` to explain that separate timeouts per-client won't work with FreeTDS <1.0rc5.

With newer versions, we can simply call the per-client `dbsetopt` function, so they will be all separate. However, with older versions all previously set timeout values will be overwritten each time a new client is created, since we have to fall back to the global `dbsettime` call. 

Also, I've rebased and squashed my commits to keep the history clean.

attilahorvath:
I don't really know why the AppVeyor test failed, it was succeeding with the same code before I rebased my branch.

Is there any way to retry it? I couldn't find a way to do that.

metaskills:
> Is there any way to retry it? I couldn't find a way to do that.

Since Appveyor is a paid service, they do not expose that rebuild without being logged in. I just hit it for you and will watch this today.

attilahorvath:
Thanks, it seems to have passed on the second try.

metaskills:
Thanks so much for this work! I've merged it in and it will be out in v2.1.0.pre3 a bit later today.

PR #351, opened 2017-03-15
FreeTDS: Link libgcc statically

### LINKED ISSUE #349: Undefined method connect for TinyTds::Client In Windows
I 'm trying to install tiny_tds in my windows machine.

This is the result I get when I `tsql -C `
```
[TinyTds][v1.1.0][tsql]: C:/Ruby22/lib/ruby/gems/2.2.0/gems/tiny_tds-1.1.0-x86-
ingw32/exe/tsql.EXE
```

This is the error message I get when I try executing  
`TinyTds::Client.new(username: "xx", password: "xxx", host: "xxxx", database: "xxxxxx")
`

 ```

NoMethodError: undefined method `connect' for #<TinyTds::Client:0x2eda828>
        from C:/Ruby22/lib/ruby/gems/2.2.0/gems/tiny_tds-1.1.0-x86-mingw32/lib/t
iny_tds/client.rb:53:in `initialize'
        from (irb):12:in `new'
        from (irb):12
        from C:/Ruby22/bin/irb:11:in `<main>'

```

orgads:
Fixes #349

### DIFF
--- a/ext/tiny_tds/extconf.rb
+++ b/ext/tiny_tds/extconf.rb
@@ -251,6 +251,16 @@ def define_freetds_recipe(host, libiconv, libssl, gnutls)
 
     class << recipe
 
+      def configure
+        cross = enable_config("cross-build")
+        if cross
+          prev_cc = ENV['CC']
+          ENV['CC'] = consolidated_host(gcc_cmd) + ' -static-libgcc'
+        end
+        super
+        ENV['CC'] = prev_cc if cross
+      end
+
       def install
         super_value = super
         # Install binstub target binaries.


orgads:
Ok, I built it and can confirm it solves the issue.

orgads:
Ah, wait. It works with devkit but not with rake-compiler-dock. Let me do some more tests.

orgads:
Ok, now it works both on cross-compilation and on devkit.

I tried many ways to pass the flags without overriding the CC variable, but couldn't. See [This thread](https://www.mail-archive.com/mingw-w64-public@lists.sourceforge.net/msg09680.html) for another user that faced this problem...

metaskills:
I'll try this out tonight. If it works, do we think a 1.2.1 release with it?

aharpervc:
FWIW I manually built the gem with this patch, and things seem to work okay for me 👍

metaskills:
Thanks! I just pushed v1.3.0 to rubygems.

Issue #216, opened 2015-08-07
Does tiny_tds support executing 'PRINT 'Hello World'' statement?


zhongxiao37:
I wonder if tiny_tds could support executing 'PRINT 'Hello World'' statement? How could I make it? Thanks in advance.

```
result = Client.execute("PRINT 'Hello World'").do
p result
```

metaskills:
I am almost certain that `PRINT` output should not go to the results stream. It could go to STDOUT but that sounds wrong too. Can you do some research on the following:

1. Does FreeTDS/DBLIB contain any interface to capture these?
2. Is there any history with this type of request to Rails core for MySQL, PostgreSQL, and/or SQLite?

I did some quick research on the first point and could not find anything.

metaskills:
Started this branch https://github.com/rails-sqlserver/tiny_tds/tree/print_feature

aharpervc:
This was added with #352 and released with 2.1.0

PR #352, opened 2017-03-22
Info Messages

aharpervc:
### Problem

I work on a codebase that makes extensive use of info messages (eg, raiserror with a serverity level <= 10, such as `raiserror('hello world', 10 -1)`). I need to receive these info messages in ruby code, but until now TinyTds did not expose an API for this.

### Solution

See updated iteration below

<details>
I have improved TinyTDS's to handle info messages, and introduced an experimental API to expose messages to users. This PR includes an improvement to the C code that handles errors with severity <= 10, and Ruby code that exposes messages to users.

The current philosophy of this project is to handle messages like errors. This PR does not change that perspective. Info messages are simply instances of `TinyTDS::Error`s but instead of throwing an exception, they're channeled through the new API. There is no pressing need to refactor the message handling part of TinyTDS. My PR here is an intentionally small diff on top of the existing codebase to fix my specific problem.

Here's what the current experimental iteration new API looks like for users:

```ruby
opts = { hostname: ..., username: ..., password: ... }

# optionally pass a block to the initializer
# it will be invoked for messages not associated with an #execute call
client = TinyTds::Client.new opts do |m|
  puts "connection message: #{m.inspect}"
end

result = @client.execute 'use demo_database; exec msg_test; use master;'

# optionally pass a block to a new #receive_message method
# it will be invoked for messages associated with these results
result.receive_message do |m|
  puts "result message: #{m.inspect}"
end

# unchanged: normal result record access
each_opts = { cache_rows: false }
result.each_with_index(each_opts) do |record, i|
  puts "record: #{i} #{record.inspect}"
end
```

This sample program calls the following test sp and prints the resulting output:

```sql
create procedure msg_test
as
begin
	
	raiserror('one', 10, -1)
	raiserror('two', 10, -1)
	
	select 1 x
	
	raiserror('three', 10, -1)
	raiserror('four', 10, -1)
	
	select 2 x
	
	raiserror('five', 10, -1)
	raiserror('six', 10, -1)

end
go
```

```
$ ruby .\tiny_test.rb
connection message: #<TinyTds::Error: Changed database context to 'master'.>
connection message: #<TinyTds::Error: Changed language setting to us_english.>
result message: #<TinyTds::Error: Changed database context to 'demo_database'.>
result message: #<TinyTds::Error: one>
result message: #<TinyTds::Error: two>
record: 0 {"x"=>1}
result message: #<TinyTds::Error: three>
result message: #<TinyTds::Error: four>
record: 1 {"x"=>2}
result message: #<TinyTds::Error: five>
result message: #<TinyTds::Error: six>
result message: #<TinyTds::Error: Changed database context to 'master'.>
```

### Ergonomics & Open Questions

1. It's kind of weird to have an `Error` object that you don't plan on using to raise an exception with, but I think it makes sense in this situation.
    - I considered making `TinyTDS::Message < Error`, but it seemed needlessly redundant.

2. There an question of whether info messages are more like results or more like events.
    - I prefer the event-like model, because it is (relatively) simple to implement a fire-and-forget model from C, and it also more easily enables interleaving messages & results.
    - The results-like model would look like having a `client.messages` array, where new messages are pushed on as they are received. This could be mutable by both the user and C code, so the user could do something like `client.messages.pop`. But ugh, more state management.

3. Fundamentally, messages are coming from the "connection" context, not the "result" context. Making it easy to get info messages "from results" is a fake abstraction.
    - My current thinking is that it's useful to do this, but it does require some additional ruby code in this PR. Basically, a reference to the most recent results object is kept, and if present, messages are sent to it. Otherwise, messages are handled by the client object. I think it's a lot more useful to receive messages in the context of loading result records, as they often go together. Also, in my current use of TinyTDS is, the connection is instantiated in one section of the code but `#execute` is called in another place. If the ONLY way to get messages is right next to `TinyTDS::Client.new`, that's going to result in me doing awkward gymnastics to get results & messages together. However with the current experimental API, I can get messages right next to results.
    - This is probably the biggest thing requiring feedback. I've not fully decided my own opinion on this yet, either.

### Things to review

What's the feeling on the following:

* Keeping an object instance reference around when the db client is set up (that is `VALUE TinyTdsInstance;` in C)
* Keeping a reference to the most recent results object around, so we can forward messages to it?
* Forwarding messages to the most recent results at all?
* Passing a block to the client initializer?

</details>

### DIFF
--- a/lib/tiny_tds/client.rb
+++ b/lib/tiny_tds/client.rb
@@ -10,6 +10,7 @@ module TinyTds
     }
 
     attr_reader :query_options
+    attr_reader :message_handler
 
     class << self
 
@@ -37,6 +38,12 @@ module TinyTds
       if opts[:dataserver].to_s.empty? && opts[:host].to_s.empty?
         raise ArgumentError, 'missing :host option if no :dataserver given'
       end
+
+      @message_handler = opts[:message_handler] || (block_given? ? Proc.new : nil)
+      if @message_handler && !@message_handler.respond_to?(:call)
+        raise ArgumentError, ':message_handler must implement `call` (eg, a Proc or a Method)'
+      end
+
       opts[:username] = parse_username(opts)
       @query_options = self.class.default_query_options.dup
       opts[:password] = opts[:password].to_s if opts[:password] && opts[:password].to_s.strip != ''
--- a/lib/tiny_tds/client.rb
+++ b/lib/tiny_tds/client.rb
@@ -39,7 +39,7 @@ module TinyTds
         raise ArgumentError, 'missing :host option if no :dataserver given'
       end
 
-      @message_handler = opts[:message_handler] = opts[:message_handler] || (block_given? ? Proc.new : nil)
+      @message_handler = opts[:message_handler]
       if @message_handler && !@message_handler.respond_to?(:call)
         raise ArgumentError, ':message_handler must implement `call` (eg, a Proc or a Method)'
       end
--- a/lib/tiny_tds/client.rb
+++ b/lib/tiny_tds/client.rb
@@ -39,7 +39,7 @@ module TinyTds
         raise ArgumentError, 'missing :host option if no :dataserver given'
       end
 
-      @message_handler = opts[:message_handler] || (block_given? ? Proc.new : nil)
+      @message_handler = opts[:message_handler] = opts[:message_handler] || (block_given? ? Proc.new : nil)
       if @message_handler && !@message_handler.respond_to?(:call)
         raise ArgumentError, ':message_handler must implement `call` (eg, a Proc or a Method)'
       end
--- a/test/result_test.rb
+++ b/test/result_test.rb
@@ -614,8 +614,8 @@ class ResultTest < TinyTds::TestCase
 
       if sqlserver?
 
-        describe 'using :message_handler' do
-          messages = []
+        describe 'using :message_handler option' do
+          let(:messages) { Array.new }
 
           before do
             close_client
@@ -626,14 +626,16 @@ class ResultTest < TinyTds::TestCase
             messages.clear
           end
 
+          it 'has a message handler that responds to call' do
+            assert @client.message_handler.respond_to?(:call)
+          end
+
           it 'calls the provided message handler when severity is 10 or less' do
             (1..10).to_a.each do |severity|
               messages.clear
-
               msg = "Test #{severity} severity"
               state = rand(1..255)
               @client.execute("RAISERROR(N'#{msg}', #{severity}, #{state})").do
-
               m = messages.first
               assert_equal 1, messages.length, 'there should be one message after one raiserror'
               assert_equal msg, m.message, 'message text'
@@ -644,15 +646,54 @@ class ResultTest < TinyTds::TestCase
 
           it 'calls the provided message handler for `print` messages' do
             messages.clear
-
             msg = 'hello'
             @client.execute("PRINT '#{msg}'").do
-
             m = messages.first
             assert_equal 1, messages.length, 'there should be one message after one print statement'
             assert_equal msg, m.message, 'message text'
           end
+        end
 
+        describe 'using an initializer block' do
+          let(:messages) { Array.new }
+
+          before do
+            close_client
+            @client = new_connection do |m|
+              messages << m
+            end
+          end
+
+          after do
+            messages.clear
+          end
+
+          it 'has a message handler that responds to call' do
+            assert @client.message_handler.respond_to?(:call)
+          end
+
+          it 'calls the provided message handler when severity is 10 or less' do
+            (1..10).to_a.each do |severity|
+              messages.clear
+              msg = "Test #{severity} severity"
+              state = rand(1..255)
+              @client.execute("RAISERROR(N'#{msg}', #{severity}, #{state})").do
+              m = messages.first
+              assert_equal 1, messages.length, 'there should be one message after one raiserror'
+              assert_equal msg, m.message, 'message text'
+              assert_equal severity, m.severity, 'message severity' unless severity == 10 && m.severity.to_i == 0
+              assert_equal state, m.os_error_number, 'message state'
+            end
+          end
+
+          it 'calls the provided message handler for `print` messages' do
+            messages.clear
+            msg = 'hello'
+            @client.execute("PRINT '#{msg}'").do
+            m = messages.first
+            assert_equal 1, messages.length, 'there should be one message after one print statement'
+            assert_equal msg, m.message, 'message text'
+          end
         end
 
         it 'must not raise an error when severity is 10 or less' do
--- a/test/result_test.rb
+++ b/test/result_test.rb
@@ -642,6 +642,17 @@ class ResultTest < TinyTds::TestCase
             end
           end
 
+          it 'calls the provided message handler for `print` messages' do
+            messages.clear
+
+            msg = 'hello'
+            @client.execute("PRINT '#{msg}'").do
+
+            m = messages.first
+            assert_equal 1, messages.length, 'there should be one message after one print statement'
+            assert_equal msg, m.message, 'message text'
+          end
+
         end
 
         it 'must not raise an error when severity is 10 or less' do
--- a/test/result_test.rb
+++ b/test/result_test.rb
@@ -637,7 +637,7 @@ class ResultTest < TinyTds::TestCase
               m = messages.first
               assert_equal 1, messages.length, 'there should be one message after one raiserror'
               assert_equal msg, m.message, 'message text'
-              assert_equal severity, m.severity, 'message severity'
+              assert_equal severity, m.severity, 'message severity' unless severity == 10 && m.severity.to_i == 0
               assert_equal state, m.os_error_number, 'message state'
             end
           end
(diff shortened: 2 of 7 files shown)


aharpervc:
Looking into test failures...

aharpervc:
I can replicate the 3 test failure, but still not sure why they're happening

metaskills:
Awhile back ago I worked on a PRING logger. See here, those branch has been deleted:
https://github.com/rails-sqlserver/tiny_tds/issues/216

IIRC, the solution was to assign a delegate object after instantiating the client. From there it was a matter of listening for the right message level and passing that string to the object. I think I decided to use an IO style object that responded to write, but again, can't remember. 

I do remember it was easy to do. I wonder if there are synergies to both needs here and a different way to go about it?

aharpervc:
Assigning an object after instantiation means that info messages during connection, such as default database and locale, will be dropped. This is because `connect` is called at the bottom of `initialize`. So there's no space between initialization and connection to separately hook up a message handler. Therefore there is a requirement to pass "something" into the initializer. I chose a block since it is backwards compatible and somewhat idiomatic.

But then, you realize that in real life, it's most useful to get messages in the context of the currently executing statement. Therefore, I set up an API to make that simple.

This is basically the API ergonomics question I originally posed... what is the right level of abstraction, what is the right way to receive messages, what "feels right", etc.

aharpervc:
Okay, I have an updated revision to talk about. It is simpler and I think closer to where we want to be. It ought not impact anything that's already going on and is almost entirely additive.

This approach of attaching a message handler to the connection object to receive error objects is how it's done elsewhere:

* [tedious](http://tediousjs.github.io/tedious/api-connection.html#event_infoMessage)
* [ADO.NET](https://msdn.microsoft.com/en-us/library/a0hee08w)

The API is:

```ruby
opts = { hostname: ..., username: ..., password: ... }

# optionally pass a block to the initializer
# it will be invoked for all messages
client = TinyTds::Client.new opts do |m|
  puts "connection message: #{m.inspect}"
end

# OR supply a :message_handler option on init
# client = TinyTds::Client.new host: 'localhost', message_handler: &method(:my_message_handler)

result = @client.execute 'use demo_database; exec msg_test; use master;'

# unchanged: normal result record access
result.each_with_index do |record, i|
  puts "record: #{i} #{record.inspect}"
end
```

```
$ ruby .\tiny_test.rb
connection message: #<TinyTds::Error: Changed database context to 'master'.>
connection message: #<TinyTds::Error: Changed language setting to us_english.>
result message: #<TinyTds::Error: Changed database context to 'demo_database'.>
result message: #<TinyTds::Error: one>
result message: #<TinyTds::Error: two>
record: 0 {"x"=>1}
result message: #<TinyTds::Error: three>
result message: #<TinyTds::Error: four>
record: 1 {"x"=>2}
result message: #<TinyTds::Error: five>
result message: #<TinyTds::Error: six>
result message: #<TinyTds::Error: Changed database context to 'master'.>
```

For situations where you want to buffer the messages or read them in the context of statement execution, this is still possible, you just have to do it yourself. For example: 

```ruby
messages = []
client = TinyTds::Client.new opts do |m|
  messages << m
end

result = @client.execute 'exec whatever'

result.each_with_index do |record, i|
  puts "record: #{i} #{record.inspect}"
  puts messages.shift # or whatever
end
```

aharpervc:
Not positive, but the test failure might be https://github.com/rails-sqlserver/tiny_tds/blob/master/test/result_test.rb#L566

aharpervc:
I'm stumped on this test failure. It appears like severity 10 (and... _just_ 10) fail to come through.

For diagnostics, I'm printing out the raw sql & the received severity at the top of `tinytds_msg_handler` and then changed my `calls the provided message handler when severity is 10 or less` test to be `(9..11)`:

```c
  // right after: int tinytds_msg_handler in client.c
  char str[20];
  sprintf(str, "severity '%d'", severity);
  rb_warn(str);
```

and the output running just that test (`state` param is randomized):

```
"RAISERROR(N'Test 9 severity', 9, 203)"
/mnt/c/Users/aharper/workspace/tiny_tds/test/result_test.rb:639: warning: severity '9'
"RAISERROR(N'Test 10 severity', 10, 183)"
/mnt/c/Users/aharper/workspace/tiny_tds/test/result_test.rb:639: warning: severity '0'
"RAISERROR(N'Test 11 severity', 11, 221)"
/mnt/c/Users/aharper/workspace/tiny_tds/test/result_test.rb:639: warning: severity '11'
```

Any ideas @metaskills or anyone else?

aharpervc:
Not sure why appveyor timed out (might be spurious). However, there's still a problem even though the tests are green, and that is that in practice, the ruby process is silently crashing sometimes when forwarding info messages to my rails app. I'm not 100% positive why this is, but I think it may be due to the C code transferring control back to ruby (via `rb_funcall`) during a "blocking" freetds operation. There's a comment about this regarding error handling, and it's why `nonblocking_error` exists. I have another "option" to implement info messages that re-uses that data structure, that does not cause ruby to silently crash (yay) but also results in a few test failures (boo). I'll push it for discussion purposes, though, since it's moving this process forward.

aharpervc:
Looks like it was a false negative after all. I'm already testing this branch in a rails app and everything seems to work as you'd hope. Ready for review @metaskills

metaskills:
NOTED! Will be on it very soon.

aharpervc:
Turns out this logic also works for `print` statements, so added a new test for that. Also, rebased on latest master.

@metaskills ping

test/result_test.rb
```diff
@@ -614,6 +614,47 @@ class ResultTest < TinyTds::TestCase
 
       if sqlserver?
 
+        describe 'using :message_handler' do
+          messages = []
```
metaskills: Could this be a `let(:messages) { Array.new }`

lib/tiny_tds/client.rb
```diff
@@ -37,6 +38,12 @@ def initialize(opts = {})
       if opts[:dataserver].to_s.empty? && opts[:host].to_s.empty?
         raise ArgumentError, 'missing :host option if no :dataserver given'
       end
+
+      @message_handler = opts[:message_handler] || (block_given? ? Proc.new : nil)
```
metaskills: I don't see anywhere where we allow a block to initialize nor is it tested. Can we simplify this to:

```ruby
@message_handler = opts[:message_handler]
```

test/result_test.rb
```diff
@@ -614,6 +614,47 @@ class ResultTest < TinyTds::TestCase
...
+              assert_equal state, m.os_error_number, 'message state'
+            end
+          end
+
+          it 'calls the provided message handler for `print` messages' do
+            messages.clear
+
+            msg = 'hello'
+            @client.execute("PRINT '#{msg}'").do
+
```
metaskills: Persnickety, can we remove new/empty lines in the it block?

lib/tiny_tds/client.rb
aharpervc: Being able to pass a block to the initializer is feature of this branch. However, there's a bug here that's fixed in the latest push.

test/result_test.rb
aharpervc: Done

test/result_test.rb
aharpervc: Done

lib/tiny_tds/client.rb
metaskills: Can this PR focus on explicit options only first. We can then open up init with a block later.

aharpervc: Well... it *could*, but why? It's very useful, and the primary way I've been using this feature. It's much more elegant and idiomatic than passing in an opt, and the cost of this addition is one ternary expression. Doesn't really make sense for that to be separate imo.

metaskills: You caught me... I'd like to not have it at all :) I have an aversion to adding features (any feature) that is superfluous. No where in the tests is this used, so it would be easy to remove later by mistake. But rather than test it to counter that, it is not needed, pure sugar. If we leave it in, on day someone will want the yield to be moved to a different spot for some other reason. I'll have to remind them of this conversation and the circle will go... basically, this is free as in 🐶 and I (the royal we too) would or should only care and feed the things that are truly needed. I just don't see the need here.

aharpervc: You missed that I had already added tests for passing a block 😈.

I suppose it's theoretically possible we might want to move it in the future, and if so, what does that matter? Library APIs sometimes change.

However, with this message handling architecture passing a block to the initializer is equivalent to needing to pass in the opt, since the `call`-able is itself passed into C space to be directly invoked. There is not currently any other opportunity to get something into that space safely (that I know of) except the initializer.

Finally, I think that removing it because it's pure sugar is unwise. Fluid ergonomics in libraries is the difference between a library that you enjoy using and a library that you are irritated to use. We all enjoy using libraries that are nice to use, so why wouldn't we also want to be nice to use?

aharpervc:
Okay, I removed the initializer block syntax.

aharpervc:
@metaskills ping

metaskills:
Thanks! Traveling this week... so looking toward this on the weekend.

metaskills:
Working on releasing v2.1.0.pre4 tonight. Keep an eye out and thank you!!!

aharpervc:
Woo, excellent! Thank you!

PR #368, opened 2017-07-15
Add compat with RubyInstaller-2.4.1

larskanis:
It adds gem metadata for automatic installation of the freetds package
which is part of the official MSYS2/MINGW repository now. This makes
installation of the tiny_tds source gem as easy as installation of the
binary windows gem, but works for RubyInstaller-2.4+ only.
The freetds MINGW package is maintained here:
https://github.com/Alexpux/MINGW-packages/tree/master/mingw-w64-freetds

The Appveyor tests still work with our ports build and doesn't test
against the new freetds package, so far.

This disables the perl hack for OpenSSL, when the perl version is
recent enough. Otherwise mixing git's shell commands into the path breaks
MSYS2 builds, as used by RubyInstaller-2.4.

This also removes lib directories from DLL search path, because they
don't contain any DLLs.

The update of libiconv is required for compat with MSYS2.

The patch file is derived from
https://github.com/Alexpux/MINGW-packages/tree/master/mingw-w64-freetds


tasks/ports/freetds.rb
```diff
@@ -23,7 +23,7 @@ def configure_defaults
       end
 
       if windows?
-        opts << '--sysconfdir=C:\Sites'
+        opts << '--sysconfdir=C:/Sites'
```
coderjoe: Is this going to work? According to Msys docs, paths starting with C: are considered to be native windows paths and do not get the benefit of automatic path conversion.

tiny_tds.gemspec
```diff
@@ -19,9 +19,10 @@ Gem::Specification.new do |s|
   s.extensions    = ['ext/tiny_tds/extconf.rb']
   s.license       = 'MIT'
   s.required_ruby_version = '>= 2.0.0'
-  s.add_runtime_dependency     'mini_portile2', '~> 2.0'
+  s.metadata['msys2_mingw_dependencies'] = 'freetds'
```
coderjoe: I can't tell you how happy I am that this is becoming a feature. :)

appveyor.yml
```diff
@@ -3,6 +3,7 @@ init:
   - SET PATH=C:\MinGW\msys\1.0\bin;%PATH%
   - SET RAKEOPT=-rdevkit
   - SET TESTOPTS='-v'
+  - SET MAKE=make V=1
```
coderjoe: Until today I did not know about this. :+1:

tasks/ports/freetds.rb
larskanis: `C:/Sites` is a native Windows path. All windows versions accept forward slash as path separator (try `dir "C:/Sites"`). I generally avoid backslash as path separator even on windows, because it is used for escaping as well. See also: https://github.com/oneclick/rubyinstaller2/wiki/For-gem-developers#path-separator

In this particular case I got an error while freetds build about a missing directory `C:Sites`, where the backslash was obviously interpreted as escape character.

tiny_tds.gemspec
larskanis: This is described here: https://github.com/oneclick/rubyinstaller2/wiki/For-gem-developers#msys2-library-dependency

larskanis:
>  Which versions of windows/architectures has this already been tested against?

On my netbook with Windows-10 64bit and on [Appveyor](https://ci.appveyor.com/project/rails-sqlserver/tiny-tds/build/2.0.0.710) with Windows 2012 R2.

tiny_tds.gemspec
coderjoe: Yep! already read it and I'm just super excited for it.

tasks/ports/freetds.rb
coderjoe: Well, to clarify my understanding is that the system path separator on Windows is "\" but almost all APIs support implicit normalization to the system path separator. Common exceptions to this rule are situations where you're using a software set which may be using its own file handling and relying on windows path separators or explicitly disabling normalization in order access previously invalid paths which are legal in NTFS or paths which extend beyond the 259 character boundary. In most cases this doesn't matter as long as you always use the system path separator and not the alt separator.

That said in this case you're absolutely right, it should be fine and since it was a bug it's a good fix. I just wasn't sure how it would come out the other end after passing through msys2 as it's entirely new to me. :)

Thank you very much for the clarification. :+1:

appveyor.yml
```diff
@@ -42,8 +43,8 @@ environment:
   TINYTDS_UNIT_AZURE_PASS:
     secure: fYKSKV4v+36OFQp2nZdX4DfUpgmy5cm0wuR73cgdmEk=
   matrix:
-    - ruby_version: "23-x64"
-    - ruby_version: "23"
+    - ruby_version: "24-x64"
+    - ruby_version: "24"
```
metaskills: Nice, I like that we now bracket 2.2 to 2.4

larskanis:
>    What is your estimate on how long we support RubyInstaller < 2.4?

Shrug. I tend to support stable versions up to 5 years, because I have to manage many different systems and hate to come into a dependency hell for each and every update. But this isn't my project...

>    Does 2.4 adoption seem good and is it easy for people to upgrade?

Good question. Rubyinstaller-2.4 has around 3500 downloads per day, similar to previous releases. The upgrade [changes more than previous releases did](https://rubyinstaller.org/2017/05/25/rubyinstaller-2.4.1-1-released.html). But a fresh installation is even a bit easier than previous releases.

>    Assuming we do go 2.4+ sometime, can you help yank out the old stuff?

I don't want to promise anything.

coderjoe:
@metaskills would you mind if I merged this (I notice it has been approved) so that I can rebase another PR I have on top of it?

metaskills:
Merge away! Thanks for asking.

coderjoe:
But of course! You are the maintainer after all. :)

Issue #372, opened 2017-07-26
Host FreeTDS binaries on github?


l8nite:
I filed an issue against the FreeTDS github directly too: https://github.com/FreeTDS/freetds/issues/135

We're having issues with FTP and network acls, and before we fork this gem and throw FreeTDS binaries in an S3 bucket or something, we are wondering if it would be possible to have a more "officially supported" installation that relied on HTTP instead of FTP for fetching the binary releases.

Ideally, FreeTDS distributes their binary releases via HTTP and tiny_tds switches to that location. Alternatively, tiny_tds hosts the binary releases it currently relies on via its own github page and references those?

coderjoe:
When you say hosting binaries, do you mean the FreeTDS versioned tarballs?

l8nite:
I mean the versioned tarballs, sorry I should have been more clear about that.

coderjoe:
No need to apologize. :)

I can empathize with your pain. We have had issues with the official FreeTDS FTP before. It was down for a short while which prevented tiny_tds builds from completing.

I can only speak from my own experience and situation, and with the limited time I have available to dedicate to open source work I would not want to take on the responsibility of packaging and distributing another project's source under the umbrella of tiny_tds.

If we could find someone with the ability and incentive to get this going, I'd probably refer them directly to the FreeTDS project so they could roll it into the FreeTDS github project where it would be infinitely more valuable.

I'd love to hear any other opinions.

metaskills:
Having had issues with FreeTDS being available too, I think this issue is a valid one. But, it should be made higher up to the FreeTDS group and distro mirrors they use. TinyTDS is one of many projects that use FreeTDS and them doing the right thing at the source should be the best for all of us. Closing this issue here. Maybe suggest taking it to one of these two places.

* http://lists.ibiblio.org/mailman/listinfo/freetds
* https://github.com/FreeTDS/freetds

PR #384, opened 2017-09-07
Use official HTTP mirrors instead of FTP

l8nite:
This is related to #372 - The FreeTDS maintainers have officially added support for HTTP downloads.


metaskills:
Thank you!

PR #390, opened 2017-10-17
Clarified ISSUE_TEMPLATE.md with windows arch instructions

### LINKED ISSUE #389: Create a debugging tool for common issues relating to loading tiny_tds.so
This gem is fairly complicated to set up since on POSIX systems it requires a fully functioning freetds install, and on windows it may or may not require different devkit based on the rubyinstaller version, as well as the architecture.

We should probably create a simple ruby script bundled with the gem `tinytds-troubleshoot` which inspects the environment and reports on potential problems to the user while also providing potential remedies.

If it is low impact enough it may even be able to be run automatically in the event of a library load failure to provide a developer instant issue resolution.

coderjoe:
This commit updates the ISSUE_TEMPLATE.md file with more debugging
instructions for users in windows and prompts users for more information
when submitting issues.

Fixes #389

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,12 +1,38 @@
-Having problems? Have you checked these first:
+## Before submitting an issue please check these first!
 
-* Have you made sure to [enable SQL Server authentication](http://bit.ly/1Kw3set)?
+* On Windows? If so, do you need devkit for your ruby install?
+* Using Ubuntu? If so, you may have forgotten to install FreeTDS first.
 * Are you using FreeTDS 0.95.80 or later? Check `$ tsql -C` to find out.
 * If not, please update then uninstall the TinyTDS gem and re-install it.
-* Using Ubuntu? If so, you may have forgotten to install FreeTDS first.
+* Have you made sure to [enable SQL Server authentication](http://bit.ly/1Kw3set)?
 * Doing work with threads and the raw client? Use the ConnectionPool gem?
 
-If none of these help. Please make sure to report:
+If none of these help. Please fill out the following:
 
-* What platform you are on. Windows, Mac, Ubuntu, etc.
-* What version of FreeTDS you are using.
+## Environment
+
+**Operating System**
+
+Please describe your operating system and version here.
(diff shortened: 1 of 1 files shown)


coderjoe:
@metaskills this seems overly simplistic, but I feel like this covers the majority of the things you might want or need people to do or collect before submitting an issue.

ISSUE_TEMPLATE.md
```diff
@@ -5,8 +5,10 @@ Having problems? Have you checked these first:
 * If not, please update then uninstall the TinyTDS gem and re-install it.
 * Using Ubuntu? If so, you may have forgotten to install FreeTDS first.
 * Doing work with threads and the raw client? Use the ConnectionPool gem?
+* On Windows? If so, do you need devkit for your ruby install?
```
metaskills: Can we move this to the first line? Idea being put these in order we think are most issues first.

Can we change the first line from `Having problems? Have you checked these first:` to something like this with BOLD markdown?

**PLEASE CHECK THESE FIRST**

ISSUE_TEMPLATE.md
```diff
@@ -5,8 +5,10 @@ Having problems? Have you checked these first:
...
 
 If none of these help. Please make sure to report:
 
 * What platform you are on. Windows, Mac, Ubuntu, etc.
-* What version of FreeTDS you are using.
+* What version of TinyTDS are you using?
+* What version of FreeTDS you are using?
```
metaskills: Maybe we have commands to run here after the question. Like:

Run `tsql -C` to see. Optionally show output.

coderjoe: These are all good ideas. I'll make these changes and update either tonight or tomorrow. 👍

coderjoe:
I've re-ordered the "check this first" section according to a completely non-scientific search of issue frequency by eyeballing the last 100 issues or so. I've made it bold and renamed the header to be more commanding to "before submitting an issue check these first!". I've also added explicit commands, where possible, to help users collect useful information about their architecture and software versions.

metaskills:
Love this new format :)

Issue #400, opened 2018-01-12
Message handler not working as expected with > 1 instance


aharpervc:
Finally getting around to using info messages, and I think I messed something up. I believe message handler references are being shared in C code, such that instantiating a second TinyTds instance with a separate message handler will overwrite the reference from the first instantiation, and things will break.

Here's a simple program to demonstrate (confirmed on ruby 2.4.1, tiny tds 2.1.0, windows 10 and ruby 2.4.3, tiny_tds 2.1.0, ubuntu 16.04)

```ruby
require 'tiny_tds'

class DB
  def initialize n
    puts "#{n} | #{object_id} | init"

    @n = n
    @client = TinyTds::Client.new host: ENV['SQL_SERVER_HOSTNAME'],
                                  username: ENV['SQL_SERVER_USERNAME'],
                                  password: ENV['SQL_SERVER_PASSWORD'],
                                  message_handler: method(:message_handler)
  end

  def message_handler e
      puts "#{@n} | #{object_id} | message_handler: #{e.message}"
  end

  def execute sql
    puts "#{@n} | #{object_id} | execute: #{sql}"

    @client.execute(sql).each_with_index(cache_rows: false, timezone: :utc) do |record, i|
      puts record
    end
  end
end


db1 = DB.new 1
db1.execute "print 'hello'"

puts

db2 = DB.new 2
db2.execute "print 'world'"

puts

# bonus points: uncomment for segfault
# db2 = nil
# GC.start(full_mark: true, immediate_sweep: true)

db1.execute "print 'I am one'"

puts

db2.execute "print 'I am two'"
```

Here's the output:

```
1 | 40909644 | init
1 | 40909644 | message_handler: Changed database context to 'master'.
1 | 40909644 | message_handler: Changed language setting to us_english.
1 | 40909644 | execute: print 'hello'
1 | 40909644 | message_handler: hello

2 | 40908180 | init
2 | 40908180 | message_handler: Changed database context to 'master'.
2 | 40908180 | message_handler: Changed language setting to us_english.
2 | 40908180 | execute: print 'world'
2 | 40908180 | message_handler: world

1 | 40909644 | execute: print 'I am one'
2 | 40908180 | message_handler: I am one # <----------- THIS IS THE PROBLEM LINE

2 | 40908180 | execute: print 'I am two'
2 | 40908180 | message_handler: I am two
```

[...]

PR #401, opened 2018-01-12
Move message_handler from a shared value to userdata

aharpervc:
- previously, subsequent instantiations of a client object would overwrite the message handler with their reference, causing messages to be handled by the latest client only
- now, the shared property is removed and each client object receives their own messages


aharpervc:
Pretty sure the CircleCi failure is unrelated.

aharpervc:
@metaskills ptal. Also, are you wanting a test that specifically covers this multi client case? It can be done (similar to the program in #400) it'll just be pretty large & awkward to cover something that's relatively hard to re-break in the future.

aharpervc:
> I have a question. Would it make any difference if you called to_proc on the Method objects returned from the method call? Not familiar with passing those objects around. That said, I'm thinking it would still be the same outcome?

I do not believe this would be a relevant change, as the apparent problem isn't the kind of object, but that there was a shared value that shouldn't be shared.

aharpervc:
Would you be willing to cut a new rubygems release for this? It'd be great to get out asap.

metaskills:
Sure, just built and pushed v2.1.1 to rubygems.

aharpervc:
💯 thanks

metaskills:
Heads up @aharpervc. I created rails-sqlserver to an org and want to add you to this team for new access control. https://github.com/orgs/rails-sqlserver/teams/contributors

Issue #184, opened 2015-01-21
ApplicationIntent:Readonly in tiny_tds?


SebastianGarin:
Is this possible? I've tried several ways of adding this string to the connection and it isn't documented.
I'm trying to connect to MSSQL high availability group. 
Any clue/idea around this will be greatly appreciated.
Thank you

metaskills:
You will have to provide more information. Do you have any links on what you are talking about?

SebastianGarin:
Hi, I want to use the connection string that is required in order to connect to a high availability group.
This string makes the following: "Connections with the "ReadOnly" property set in the connection string will be redirected to the secondary replica"

The "ApplicationIntent=ReadOnly" is added to the connection in this tab:

http://blogs.msdn.com/cfs-filesystemfile.ashx/__key/communityserver-blogs-components-weblogfiles/00-00-00-84-42-metablogapi/7610.clip_5F00_image007_5F00_thumb_5F00_3589CC7F.png

Without this I cannot connect to the MSSQL DB.

When trying to connect, the error given by tiny_tds is:

/var/lib/gems/1.9.1/gems/tiny_tds-0.6.2/lib/tiny_tds/client.rb:77:in `connect': The target database ('database') is in an availability group and is currently accessible for connections when the application intent is set to read only. For more information about application intent, see SQL Server Books Online. (TinyTds::Error)
        from /var/lib/gems/1.9.1/gems/tiny_tds-0.6.2/lib/tiny_tds/client.rb:77:in `initialize'
        from MSSQL.rb2:5:in `new'
        from MSSQL.rb2:5:in `<main>'

Again, I couldn't find any documentation regarding this.

Thanks,

metaskills:
Very interesting. Thanks for the details. I did some Googling on "freetds application intent read only" and found some mentions that you have to use a user that only has readonly access. Is this the case for you? Also, I just found this too.

https://gitorious.org/freetds/readintent-freetds/source/10a7f55ed3ac2b095910226f40768ef4a91e8e83:

If this is true, I will have to go digging in the FreeTDS code and make some changes. This will take me a few weeks since I am working on the latest 4.2 adapter now.

SebastianGarin:
Yes, I'm a read only user on that db. Should there be a way of configuring freetds and use it as a "DSN" via tinytds?
Thank you for your feedback Ken

metaskills:
> configuring freetds and use it as a "DSN" 

[...]

PR #407, opened 2018-05-01
Add support for "application intent read only"

aharpervc:
Add support for "Application Intent Read Only" (https://github.com/rails-sqlserver/tiny_tds/issues/184). The use case is, when connecting to an availability group listener that has read only secondary replicas, the app can request to be connected to a replica rather than the primary.

Other than confirming that there's no regressions and the bit can be set, I'm not sure exactly how to write tests for this.... setting up multiple sql server docker containers just to set up an AG just to test this seems like overkill.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,7 @@ Creating a new client takes a hash of options. For valid iconv encoding options,
 * :contained - Pass true to signal that you are connecting with a contained database user.
 * :use_utf16 - Instead of using UCS-2 for database wide character encoding use UTF-16. Newer Windows versions use this encoding instead of UCS-2. Default true.
 * :message_handler - Pass in a `call`-able object such as a `Proc` or a method to receive info messages from the database. It should have a single parameter, which will be a `TinyTds::Error` object representing the message. For example:
+* :application_intent - When connecting to an Availability Group, pass `:read_only` to connect to a read replica (Note: the `:database` option must also be set to a database joined to the AG that has a read replica). Defaults to `nil`.
 
 ```ruby
 opts = ... # host, username, password, etc
--- a/ext/tiny_tds/client.c
+++ b/ext/tiny_tds/client.c
@@ -3,9 +3,9 @@
 
 VALUE cTinyTdsClient;
 extern VALUE mTinyTds, cTinyTdsError;
-static ID sym_username, sym_password, sym_dataserver, sym_database, sym_appname, sym_tds_version, sym_login_timeout, sym_timeout, sym_encoding, sym_azure, sym_contained, sym_use_utf16, sym_read_only_intent, sym_message_handler;
+static ID sym_username, sym_password, sym_dataserver, sym_database, sym_appname, sym_tds_version, sym_login_timeout, sym_timeout, sym_encoding, sym_azure, sym_contained, sym_use_utf16, sym_message_handler, sym_application_intent;
 static ID intern_source_eql, intern_severity_eql, intern_db_error_number_eql, intern_os_error_number_eql;
-static ID intern_new, intern_dup, intern_transpose_iconv_encoding, intern_local_offset, intern_gsub, intern_call;
+static ID intern_new, intern_dup, intern_transpose_iconv_encoding, intern_local_offset, intern_gsub, intern_call, intern_read_only;
 VALUE opt_escape_regex, opt_escape_dblquote;
 
 
@@ -307,7 +307,7 @@ static VALUE rb_tinytds_identity_sql(VALUE self) {
 
 static VALUE rb_tinytds_connect(VALUE self, VALUE opts) {
   /* Parsing options hash to local vars. */
-  VALUE user, pass, dataserver, database, app, version, ltimeout, timeout, charset, azure, contained, use_utf16, read_only_intent;
+  VALUE user, pass, dataserver, database, app, version, ltimeout, timeout, charset, azure, contained, use_utf16, application_intent;
   GET_CLIENT_WRAPPER(self);
 
   user = rb_hash_aref(opts, sym_username);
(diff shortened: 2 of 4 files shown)


y310:
@metaskills Hi, I think this change is needed to use recently released read-only replicas on Azure SQL DB.
Is there any problems to prevent from this PR get progress?

https://docs.microsoft.com/en-us/azure/sql-database/sql-database-read-scale-out

aharpervc:
> Is there any problems to prevent from this PR get progress?

The "problem" is that I still need to get a test lab set up to confirm this actually works as intended. I haven't had the capacity to revisit that since writing this commit.

It actually would be very helpful if you can test this on your end (check out this branch, build your own copy of the gem, connect to read scale replicas, do the appropriate diagnostics to confirm you're getting routed to the correct replica and otherwise everything seems correct).

ryu1kn:
We wanted to use read-only intent to connect to our on-premise SQL Server database cluster.

I wrote a snippet to check whether the query is executed on the intended server, and confirmed the SQL is run on the right server. The catch is, I had to specify `contained: true` as well to specify the database upon opening a connection.

https://gist.github.com/ryu1kn/a76694c42432cfd62e9fdf16cb2f5c88

ryu1kn:
Without this PR, we cannot tell FreeTDS to use read-only intent from our Ruby code. One way to get around this is to specify it in [`freetds.conf` file](https://www.freetds.org/userguide/freetdsconf.htm) like this.

```
[my-db]
        host = mydb.domain.com
        port = 1433
        tds version = 7.3
        read-only intent = yes
```

Then specify `my-db` in `host` property when instantiating TinyTDS.

```ruby
client = TinyTds::Client.new(
  host: 'my-db',
  username: 'db_username',
  password: 'db_password',
  database: 'db_databasename',
  contained: true,
  message_handler: Proc.new { |m| puts m.message }
)
```

aharpervc:
Alright, I had some time to revisit this branch. I have an AG lab set up locally to verify that setting the new option does indeed work as expected.

There's some nuance to this:

- per @saurabh500's comment [here](https://github.com/rails-sqlserver/tiny_tds/issues/184#issuecomment-148497780), support was only added for this bit in TDS 7.4. Therefore, I added a warning if you set `:intent_read_only` and are using some prior version
- per @ryu1kn's comments, the reason why you have to set `:contained` is that that the code path that sets the login database is only processed when you also pass `:azure` or `:contained`. Not sure why it was done that way ¯\\_(ツ)_/¯. However, I have updated that code path to read in the provided default database when passing `:intent_read_only` as well
- This required also updating the version table to add a 7.4 entry. I also updated the default, but that might not be desirable and we can back it out if so. That'd mean that if you want to use `:intent_read_only`, you also need to initialize with `:tds_version => '7.4'`
- Since we don't have an AG in the testing setup, there's not really a good way to verify that things all work. However I have still added an test that you can opt-in to running, by setting `TINYTDS_READ_REPLICA_SERVER_NAME`. Then run it like `TINYTDS_READ_REPLICA_SERVER_NAME=mssql_2 m test/client_test.rb:74` and it'll verify that TinyTDS got redirected to the read replica you told it to expect. This works for me locally and if anyone that can run these tests in their AG, that'd be great as well.

aharpervc:
This should maybe change from a bool to a symbol enumeration, to support Application Intent Read Write https://docs.microsoft.com/en-us/sql/database-engine/availability-groups/windows/secondary-replica-connection-redirection-always-on-availability-groups?view=sql-server-ver15

aharpervc:
> This should maybe change from a bool to a symbol enumeration, to support Application Intent Read Write

The config option is now called `:application_intent` and supports a value of `:read_only` to set the corresponding login bit. This more closely matches other libraries & microsoft's documentation.

Building & testing locally with an AG w/ read replicas confirms that this still works (ie, tiny tds is talking to the replica host when `application_intent: :read_only` is provided).

> Next step is probably to get a consensus one way or the other about the 7.4 configuration, rework the branch commit messages, and get a code review from anyone available.

👆 this is still the case

gisborne:
It sounds like this work is done. Why is this PR still open?

aharpervc:
> It sounds like this work is just about done. Why is this issue still open?

Historically on this project we have released features in accordance with needs from users. At my company, we decided to choose a replication technology besides Availability Groups. Therefore, this feature ended up unrequired for us and I suspended working on the PR. Also, as you can see there were no additional responses to my request for consensus about next steps.

However, the PR has remained open in case someone else voices a need for this feature or would like to contribute to it.

david-alex-wynn:
Hello,

I am fairly green to the open source community, and am not sure if this is the proper way to communicate this, but we could really use this feature. I pulled in your branch and tested it out. Everything appears to be working exactly as we would expect. Is there anything I can do to help get this merged?

rud (approved):

metaskills:
Closing due to age and cleaning up my PR inbox. Feel free to open, talk, etc.

dhockett-dmp:
I'll jump in and say this would be highly valuable to us.

With regards to the version default, as this seems to be a fairly specific use case, I think it's reasonable that gem consumers would need to specify the new version in order to access the feature. So if there are concerns with bumping the default, the additional requirement of directly specifying `:tds_version=>7.4` doesn't sound like a blocker to me.

Is there anything else in particular that's required to resurrect this?

elitenomad:
Any idea why this PR is closed and not merged to master ? Don't see application-intent related code on master branch ?

aharpervc:
> > It sounds like this work is just about done. Why is this issue still open?
> 
> Historically on this project we have released features in accordance with needs from users. At my company, we decided to choose a replication technology besides Availability Groups. Therefore, this feature ended up unrequired for us and I suspended working on the PR. Also, as you can see there were no additional responses to my request for consensus about next steps.
> 
> However, the PR has remained open in case someone else voices a need for this feature or would like to contribute to it.

I explained it above, still true even today. If anyone has the desire to work on this, they are welcome/encouraged too! To get started, you can fork/copy my original attempt and make a new PR, and then we can continue some of the old discussion there.

PR #420, opened 2018-10-31
Add documentation about linking in Docker

### LINKED ISSUE #419: TinyTds with Docker multistage builds
## Environment

**Operating System**

Please describe your operating system and version here.
If unsure please try the following from the command line:

Linux 4124334e150d 4.9.125-linuxkit #1 SMP Fri Sep 7 08:20:28 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

Ubuntu 18.04

tiny_tds (2.1.2)

**FreeTDS Version**

freetds-1.00.92

## Description

I am building a Rails application with a multi-stage Docker build. My Dockerfile looks something like this, although we have a base image that isn't quite ubuntu 18.04 that adds some Oracle libraries we need, Passenger, and sets up nginx to run as a non-root user:
```
# Build free tds and put it somewhere easy to get to
FROM ubuntu:18.04 as tds_builder
RUN apt-get update && apt-get install -y wget gcc make && \
  wget ftp://ftp.freetds.org/pub/freetds/stable/freetds-1.00.92.tar.gz && \
  tar -xzf freetds-1.00.92.tar.gz && \
  cd freetds-1.00.92 && \
  ./configure --prefix=/home/deploy/tds && \
  make && make install && make clean

# Build all our gems
FROM ubuntu:18.04 as builder
ENV PATH="${PATH}:/home/.gem/bin"
COPY --from=tds_builder /home/deploy/tds/lib /usr/local/lib
COPY --from=tds_builder /home/deploy/tds/include/* /usr/local/include/
COPY --from=tds_builder /home/deploy/tds/etc/* /etc/freetds/
RUN apt-get update && apt-get install -y curl links g++ git ruby-dev zlib1g-dev make && \
  chown deploy:www-data /var/www
USER deploy
RUN mkdir /var/www/app && \
  git clone <ruby-app-git-url> --depth=1 --branch stable --single-branch /var/www/app && \
  cd /var/www/app && \
  gem install rake --no-document --no-ri && \
  gem install bundler --no-document --no-ri &&\
  bundle install --without development test

# Set up final container
FROM  ubuntu:18.04 as final
ENV PATH="${PATH}:${ORACLE_BASE}:/home/.gem/bin"
COPY --from=builder --chown=deploy:www-data /home/.gem /home/.gem
COPY --from=builder --chown=deploy:www-data /var/www/app /var/www/app
COPY --from=builder --chown=deploy:www-data /etc/default/nginx /etc/default/nginx
COPY --from=builder /usr/local/lib /usr/local/lib
COPY --from=builder /usr/local/include/* /usr/local/include/
COPY --from=builder /etc/freetds /etc/freetds

EXPOSE 8080
ENTRYPOINT [ "start.sh" ]
```

We have several applications we do this with, but the ones using TinyTDS encounter the following error: `Error: The application encountered the following error: libsybdb.so.5: cannot open shared object file: No such file or directory - /home/.gem/gems/tiny_tds-2.1.2/lib/tiny_tds/tiny_tds.so (LoadError)`

It seems like TinyTDS relies on something outside of the gem directory, and in addition to the free tds lib/include files when it builds its native extension that gets lost in the copy. None of the other gems we include have had this problem since we use this builder pattern successfully with our other apps.

If I add make and gcc to the final container and do a `gem install tiny_tds` everything starts to work again so there's definitely something extra being built that I'm not finding. Any advice?

mlh758:
Closes #419

I removed documentation that said TinyTDS would install FreeTDS for you since that seems to have been removed in version 2. I can remove that change if it's unwanted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -472,7 +472,7 @@ If you use a [multi stage](https://docs.docker.com/develop/develop-images/multis
 into another, lighter, container without build tools you will need to make sure you tell the OS how to find dependencies for TinyTDS.
 
 After you have built and installed FreeTDS it will normally place library files in `/usr/local/lib`. When TinyTDS builds native extensions, 
-it [already knows to look here](https://github.com/mlh758/tiny_tds/blob/master/ext/tiny_tds/extconf.rb#L31) but if you copy your app to a new container that link will be broken.
+it [already knows to look here](https://github.com/rails-sqlserver/tiny_tds/blob/master/ext/tiny_tds/extconf.rb#L31) but if you copy your app to a new container that link will be broken.
 
 Set the LD_LIBRARY_PATH environment variable `export LD_LIBRARY_PATH=/usr/local/lib` and run `ldconfig`. If you run `ldd tiny_tds.so` you should not see any broken links. Make
 sure you also copied in the library dependencies from your build container with a command like `COPY --from=builder /usr/local/lib /usr/local/lib`.
--- a/README.md
+++ b/README.md
@@ -474,7 +474,7 @@ into another, lighter, container without build tools you will need to make sure 
 After you have built and installed FreeTDS it will normally place library files in `/usr/local/lib`. When TinyTDS builds native extensions, 
 it [already knows to look here](https://github.com/rails-sqlserver/tiny_tds/blob/master/ext/tiny_tds/extconf.rb#L31) but if you copy your app to a new container that link will be broken.
 
-Set the LD_LIBRARY_PATH environment variable `export LD_LIBRARY_PATH=/usr/local/lib` and run `ldconfig`. If you run `ldd tiny_tds.so` you should not see any broken links. Make
+Set the LD_LIBRARY_PATH environment variable `export LD_LIBRARY_PATH=/usr/local/lib:${LD_LIBRARY_PATH}` and run `ldconfig`. If you run `ldd tiny_tds.so` you should not see any broken links. Make
 sure you also copied in the library dependencies from your build container with a command like `COPY --from=builder /usr/local/lib /usr/local/lib`.
 
 ## Help & Support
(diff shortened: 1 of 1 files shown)


aharpervc:
Questions raised purely from reading this readme and not closely investigating the other issue:

- if freetds is normally in /usr/local/lib, and tinytds already knows how to load from /usr/local/lib, what is the purpose of setting LD_LIBRARY_PATH to /usr/local/lib?
- when talking about how to set this up with a multistage docker build, wouldn't it be better to install to another directory than default, so your `COPY` command is more obviously useful (I get that it's copying from a prior stage to the current stage, but when it's the same folder path in both stages it's less obviously interesting)
- should the url to extconf.rb be to this repo rather than your fork?

mlh758:
Are you okay with the changes now?

coderjoe:
Just re-read it. Should we really be suggesting a complete overwrite of `LD_LIBRARY_PATH`?
We should really just be adding `/usr/local/lib` to the end of the existing path right?

Something like `export LD_LIBRARY_PATH=/usr/local/lib:${LD_LIBRARY_PATH}` maybe?

coderjoe:
Alternatively we could suggest installing FreeTDS to `/usr/lib` instead of `/usr/local/lib` and then maybe you don't have to change `LD_LIBRARY_PATH` at all?

mlh758:
`export LD_LIBRARY_PATH=/usr/local/lib:${LD_LIBRARY_PATH}` would be safer, it's just normally blank and I assumed the reader would know to preserve it if necessary. I'll change it just to be safer by default.

I don't know why free tds uses `/usr/local/lib` instead, but tiny_tds also knows to look there and I'm not sure if changing that on the installed system would cause any other weird side effects.

coderjoe:
Looks good to me. Thanks. 👍

PR #422, opened 2019-01-04
Add Docker Container for fast access/prototyping

OfftheCode:
Last time I had to create a container which would connect with sqlserver running on azure and had some problems setting it all up without to much mess. So I thought I will share my solution. Image created from this Docker file is 508MB and it's built on top of the lightweight ruby-alpine version. I think might be useful for people struggling with setting this up. I have also the Docker version for regular ruby image but it's 2 times bigger in size.


dziobaczy:
I just found this in my open PR's >,<. Do you still want that update in README?

rud (approved):
Looks valuable to me

andyundso:
Since [v3.2.0](https://github.com/rails-sqlserver/tiny_tds/releases/tag/v3.2.0), `tiny_tds` ships the necessary dependencies in the Linux gem, so you are longer required to compile FreeTDS yourself, making this PR obsolete.

PR #423, opened 2019-01-11
Update recommended minor version of FreeTDS

nzdjb:
The readme currently suggests using a minor version of FreeTDS that is quite out of date and doesn't work with OpenSSL v1.1.0. OpenSSL v1.1.0 is the version installed on most modern operating systems so this can cause confusion for users.

Support for OpenSSL v1.1.0 was added in freetds-1.00.27, however the latest version is 1.00.110. Installation of tiny_tds seems to work fine with 1.00.110.


nzdjb:
Closing this for now. There's work needed to get the build going. I'll submit separate pull requests.

PR #424, opened 2019-01-11
Fix build

nzdjb:
Due to updates to the bundler gem making it incompatible with older versions of ruby, the build is broken. This change specifies a compatible version of bundler for older rubies, allowing the build to pass again.


nzdjb:
Closing in favour of [PR#428](https://github.com/rails-sqlserver/tiny_tds/pull/428).

PR #428, opened 2019-02-18
Upgrade OpenSSL to next minor version 1.1.0j

sukeerthiadiga:
- OpenSSL 1.1.0e -> 1.1.0j
- Change travis-ci config file to pick specific version of bundler.
- Added 2.4, 2.5 and 2.6 ruby versions for testing.


sukeerthiadiga:
@metaskills by any chance can we merge this.
Credits to @nzdjb. I have pulled his code to fix the build.

aharpervc:
If you're going to spend the time to reconfigure the build matrix as a prereq for upgrading openssl, I think we should take that opportunity to drop testing for 2.1 (EOL'd 2017-03-31) and 2.2 (EOL'd 2018-03-31). Even 2.3 EOL's 2019-03-31, so I guess we can keep it for now. And then we should probably ensure that 2.4 and 2.5 are tested, and ideally 2.6 as well.

sukeerthiadiga:
> If you're going to spend the time to reconfigure the build matrix as a prereq for upgrading openssl, I think we should take that opportunity to drop testing for 2.1 (EOL'd 2017-03-31) and 2.2 (EOL'd 2018-03-31). Even 2.3 EOL's 2019-03-31, but I guess we can keep it for now. And then we should probably ensure that 2.4 and 2.5 are tested, and ideally 2.6 as well.

This totally makes sense to me. I have added latest 2.4.x, 2.5.x and 2.6.x ruby versions. We should be good now.

wpolicarpo:
I agree we should drop ruby 2.1 and 2.2 from the matrix. In the end, updating the `rvm` matrix values was enough.

nzdjb:
I'm going to close out [PR#424](https://github.com/rails-sqlserver/tiny_tds/pull/424) in favour of this.

PR #429, opened 2019-02-20
Bump version to 2.1.3

sukeerthiadiga:
This will help in getting OpenSSL to next minor version 1.1.0j


wpolicarpo:
No need to bump it now. We can bump the version when the release data arrives (it might not be just a tiny bump).

PR #431, opened 2019-02-22
 Use official HTTP mirrors instead of FTP for 1-stable

januszm:
Related PR: https://github.com/rails-sqlserver/tiny_tds/pull/384

Change FreeTDS source server from FTP to HTTP

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ $ apt-get wget
 $ apt-get install build-essential
 $ apt-get install libc6-dev
 
-$ wget ftp://ftp.freetds.org/pub/freetds/stable/freetds-1.00.21.tar.gz
+$ wget http://www.freetds.org/files/stable/freetds-1.00.21.tar.gz
 $ tar -xzf freetds-1.00.21.tar.gz
 $ cd freetds-1.00.21
 $ ./configure --prefix=/usr/local --with-tdsver=7.3
--- a/ext/tiny_tds/extconsts.rb
+++ b/ext/tiny_tds/extconsts.rb
@@ -7,9 +7,9 @@ OPENSSL_SOURCE_URI = "https://www.openssl.org/source/openssl-#{OPENSSL_VERSION}.
 
 FREETDS_VERSION = ENV['TINYTDS_FREETDS_VERSION'] || "1.00.27"
 FREETDS_VERSION_INFO = Hash.new { |h,k|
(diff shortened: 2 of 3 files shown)


januszm:
It'd be great if one of the maintainers could have a look at failing specs. Looks like the test MSSQL server is not available?

wpolicarpo:
@metaskills can you have a look here? I'm not sure if you want maintain `1.3.x` given you decided to remove `freetds` installation from the gem itself (which I agree).

aharpervc:
The http archive goes down all the time as well, unfortunately (doesn't mean we shouldn't merge this, just fyi)

januszm:
Can't we upload source files to a more reliable cloud storage or CDN? (e.g. Amazon S3, CloudFront) ?

wpolicarpo:
We can, but there are many problems with this approach:

1. Security: How to guarantee the integrity of downloaded file if we host it somewhere?
2. Cost: After all, someone would need to pay for it.
3. Maintainers would have to monitor/maintain another source of artifacts

I personally don't see any issues in merging this, but I don't think it will solve the main issue, which is managing external dependencies and probably that's why they changed it in 2.x.

In my opinion, installing `freetds` should be a step of your deploy. It's more secure and reliable if we do that by ourselves than delegating this to a gem installation.

januszm:
@wpolicarpo 

> In my opinion, installing `freetds` should be a step of your deploy. It's more secure and reliable if we do that by ourselves than delegating this to a gem installation.

I can not agree more, but it's not always easily available on certain Linux distributions, and this mechanism of downloading and installing `freetds` is just a nice "emergency" mechanism.

This still does not seem like a fully reliable solution, but, you know what they say: "something is better than nothing".

andyundso:
appears to be the same change as in #384.

andyundso:
ah I see it targets v1 instead of `main`. don't think we support this version anymore.

januszm:
It's fine, for Amazon Linux I found a better solution, however it requires some manual maintenance, maybe once a year. For those who are interested, it's all about compiling FreeTDS on an Amazon Linux X (e.g. 2023), packaging it as zip or tar.gz and hosting on S3 bucket. It can be then downloaded fast (same Amazon Data Center ideally) and extracted to target system:

```
commands:
  install_freetds:
    command: |
      aws s3 cp s3://YOUR_S3_BUCKET_HERE/freetds-al2023/freetds.tar.bz2 freetds.tar.bz2
      tar xfo freetds.tar.bz2 -C /usr/
      mv /usr/etc/{freetds.conf,locales.conf,pool.conf} /etc/
```

Issue #434, opened 2019-04-23
Please prepare tiny_tds gem for Ruby v2.6 on Windows


TheR2:
Windows 10

ERROR:  Error installing tiny_tds:
        The last version of tiny_tds (>= 0) to support your Ruby & RubyGems was 2.1.2. Try installing it with `gem install tiny_tds -v 2.1.2`
        tiny_tds requires Ruby version >= 2.0, < 2.6. The current ruby version is 2.6.1.33.

Thank you
TheR

aharpervc:
Huh, looks like the rake task might need to be updated to match the ruby versions we're testing in CI

https://github.com/rails-sqlserver/tiny_tds/blob/master/tasks/native_gem.rake#L11 lists 2.5.0, 2.4.0, 2.3.0, 2.2.2, 2.1.6, and 2.0.0 whereas https://github.com/rails-sqlserver/tiny_tds/blob/master/appveyor.yml lists 2.3 through 2.6.

I can't think of any reason for those lists to be different, to be honest.

avinashsa:
folks, what does this mean, "gem install"  still complains about this error on windows 10

will:
Yeah I can confirm that you can't install using ruby 2.6 on windows

mrambossek:
any updates - is there a workaround? thanks

aharpervc:
Did you try building the gem after updating the rake task to include 2.6 as I described above?

mrambossek:
> Did you try building the gem after updating the rake task to include 2.6 as I described above?

sorry i dont know how to do that.
i did a git checkout, and then ran "gem build tiny_tds.gemspec", which produced a tiny_tds-2.1.2.gem which i then installed via "gem install". however now i appearently dont have any of those tinytds static libraries? not sure. what are the steps necessary to get or build those tinytds libraries?

aharpervc:
> what are the steps necessary to get or build those tinytds libraries?

Please review the instructions in the readme file. If you're on Windows, the simplest way is to use docker to set up the build environment. https://github.com/rails-sqlserver/tiny_tds#compiling-gems-for-windows

mrambossek:
> Please review the instructions in the readme file. If you're on Windows, the simplest way is to use docker to set up the build environment. https://github.com/rails-sqlserver/tiny_tds#compiling-gems-for-windows

tried this, i think im on the right track, but the build fails somewhere on openssl?
`
[...]

PR #455, opened 2020-01-31
Remove old Ruby version from CI config & cross build rake task

aharpervc:
This is a prerequisite for https://github.com/rails-sqlserver/tiny_tds/issues/434 (per [comment](https://github.com/rails-sqlserver/tiny_tds/issues/434#issuecomment-485863370)). It's effectively a followup to https://github.com/rails-sqlserver/tiny_tds/pull/428.

Adding recent contributors for review.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,6 @@ env:
     - TESTOPTS="-v"
     - TINYTDS_UNIT_HOST=localhost
 rvm:
-  - 2.3.8
   - 2.4.5
   - 2.5.3
   - 2.6.1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## (unreleased)
+
+* Removed old/unused appveyor config
+* Remove old Rubies from CI & cross compile list
+* Add Ruby 2.6 to the cross compile list
+
 ## 2.1.2
 
 * Use Kernel.BigDecimal vs BigDecimal.new. Fixes #409.
--- a/README.md
+++ b/README.md
@@ -401,7 +401,7 @@ The default is true and since FreeTDS v1.0 would do this as well.
 
 ## Compiling Gems for Windows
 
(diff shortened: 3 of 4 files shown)


aharpervc:
@larskanis looks like you were also doing this work here: https://github.com/rails-sqlserver/tiny_tds/compare/master...larskanis:update-windows-binary, was there a blocker on that?

jrgns:
Can we perhaps add support for Ruby 2.7 while we're at it?

aharpervc:
> Can we perhaps add support for Ruby 2.7 while we're at it?

I don't see why not

larskanis:
@aharpervc I don't remember what interrupted me. Nevertheless you need to update rake-compiler-dock to 1.0 in order to cross build for Ruby-2.7. OpenSSL-1.1.1 should work as well. We cross build it in ruby-pg.

larskanis:
Appveyor seems to be disabled for this repository. The badge icon in the README shows the last build two years ago.

aharpervc:
Great @larskanis thanks for the tip. I removed the old appveyor config and updated rake-compiler-dock & the default OpenSSL version.

aharpervc:
The 2.7 job is failing because tiny tds can't find the freetds headers (I think...) indicated by this right before the final error:

```log
checking for sybfront.h... no
checking for sybdb.h... no
checking for tdsdbopen() in -lsybdb... no
checking for dbanydatecrack() in -lsybdb... no
```

However, I don't know why that would be. Freetds appears to install without problem, and diffing the 2.6 & 2.7 job logs that part of each job seems to be similar. @larskanis ideas?

aharpervc:
Given that 2.7 is requires additional changes, I'm going to slice that work out to another branch so we can get the other changes merged.

larskanis:
@aharpervc Couldn't you reactivate Appveyor-CI instead of dropping it? That would make sure, that the gem is properly working on Windows. Or is there a particular reason to not use Appveyor any longer?

aharpervc:
> Or is there a particular reason to not use Appveyor any longer?

I feel like I asked this question somewhere on this repo a long time ago, and there was an answer, but I can't recall what it was & I can't find my question.

I'm open to adding more CI config, especially for Windows. Whether that would be back to appveyor or some other service is an open question to me. Travis has Windows support now I believe, and CircielCI supports Linux/macOS/Windows and might be a better choice overall. Github Actions can be used for CI and also supports all the OS's.

MadsNielsen:
Hey @aharpervc is there any change we might soon see a new release of this gem with the support for 2.6 and precompiled windows gems?

aharpervc:
I hope so, although this branch has been superseded by https://github.com/rails-sqlserver/tiny_tds/pull/456

PR #469, opened 2020-05-15
Raise error if client is unable to execute statement

### LINKED ISSUE #451: Client should raise error on failed execution
## Environment
Debian (Docker)
Image `ruby:2`

I am wondering if `execute` should raise `TinyTds::Error` on failed execution?  
See more info here:
https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/717

To summarize the consumer got `false` as a return value which IMO totally unexpected.  
I wonder did you forgot to raise an error in ruby instead of returning false from the native C to the consumer?

### LINKED ISSUE #465: 2.13.pre doesn't load on win10
I on Windows 10 running ruby 2.6.3p62 (2019-04-16 revision 67580) [i386-mingw32]
I have tiny_tds (2.1.3.pre x86-mingw32) installed
When I try to execute me code I get

C:/Ruby26/lib/ruby/2.6.0/rubygems/core_ext/kernel_require.rb:54:in `require': 126: The specified module could not be found.   - C:/Ruby26/lib/ruby/gems/2.6.0/gems/tiny_tds-2.1.3.pre-x86-mingw32/lib/tiny_tds/2.6/tiny_tds.so (LoadError)

         9: from C:/Ruby26/lib/ruby/gems/2.6.0/gems/tiny_tds-2.1.3.pre-x86-mingw32/lib/tiny_tds.rb:42:in `<top (required)>'
         8: from C:/Ruby26/lib/ruby/gems/2.6.0/gems/tiny_tds-2.1.3.pre-x86-mingw32/lib/tiny_tds.rb:33:in `block in <top (required)>'
         7: from C:/Ruby26/lib/ruby/gems/2.6.0/gems/tiny_tds-2.1.3.pre-x86-mingw32/lib/tiny_tds.rb:17:in `block in <top (required)>'
         6: from C:/Ruby26/lib/ruby/gems/2.6.0/gems/tiny_tds-2.1.3.pre-x86-mingw32/lib/tiny_tds.rb:24:in `rescue in block in <top (required)>'
         5: from C:/Ruby26/lib/ruby/gems/2.6.0/gems/tiny_tds-2.1.3.pre-x86-mingw32/lib/tiny_tds.rb:34:in `block (2 levels) in <top (required)>'
         4: from C:/Ruby26/lib/ruby/gems/2.6.0/gems/tiny_tds-2.1.3.pre-x86-mingw32/lib/tiny_tds.rb:37:in `block in <top (required)>'
         3: from C:/Ruby26/lib/ruby/gems/2.6.0/gems/tiny_tds-2.1.3.pre-x86-mingw32/lib/tiny_tds.rb:43:in `block in <top (required)>'
         2: from C:/Ruby26/lib/ruby/gems/2.6.0/gems/tiny_tds-2.1.3.pre-x86-mingw32/lib/tiny_tds.rb:46:in `rescue in block in <top (required)>'
         1: from C:/Ruby26/lib/ruby/2.6.0/rubygems/core_ext/kernel_require.rb:54:in `require'
C:/Ruby26/lib/ruby/2.6.0/rubygems/core_ext/kernel_require.rb:54:in `require': cannot load such file -- tiny_tds/tiny_tds (LoadError)

I would have expected this to work.

wpolicarpo:
`TinyTDS` returns `false` when you try to execute a query and for some reason it fails (dead connection, a network intermittent error, etc). This PR changes this behavior to raise an exception instead.

This is particular useful for ORMs built on top of TinyTDS connection like Sequel or ActiveRecord. Note that this also breaks the public API and this might be undesired.

Fixes #451.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * Improve handling of network related timeouts
 * Fix error reporting when preceded by info message
+* Raise error if client is unable to execute statement
 
 ## 2.1.3
 
--- a/ext/tiny_tds/client.c
+++ b/ext/tiny_tds/client.c
@@ -296,8 +296,8 @@ static VALUE rb_tinytds_execute(VALUE self, VALUE sql) {
   REQUIRE_OPEN_CLIENT(cwrap);
   dbcmd(cwrap->client, StringValueCStr(sql));
   if (dbsqlsend(cwrap->client) == FAIL) {
-    rb_warn("TinyTds: dbsqlsend() returned FAIL.\n");
-    return Qfalse;
+    rb_raise(cTinyTdsError, "failed to execute statement");
+    return self;
   }
   cwrap->userdata->dbsql_sent = 1;
   result = rb_tinytds_new_result_obj(cwrap);
--- a/test/client_test.rb
+++ b/test/client_test.rb
@@ -132,6 +132,25 @@ class ClientTest < TinyTds::TestCase
       end
     end
 
+    it 'raises TinyTds exception when tcp socket is down' do
+      skip if ENV['CI'] && ENV['APPVEYOR_BUILD_FOLDER'] # only CI using docker
+      begin
+        client = new_connection timeout: 2, port: 1234
+        assert_client_works(client)
+        action = lambda { client.execute("waitfor delay '00:00:01'").do }
+
+        # Use toxiproxy to close the TCP socket immediately.
+        # We want TinyTds to fail to execute the statement and raise an error immediately.
+        Toxiproxy[:sqlserver_test].down do
+          assert_raise_tinytds_error(action) do |e|
+            assert_match %r{failed to execute statement}i, e.message, 'ignore if non-english test run'
+          end
+        end
+      ensure
+        assert_new_connections_work
+      end
+    end
+
     it 'raises TinyTds exception with tcp socket network failure' do
       skip if ENV['CI'] && ENV['APPVEYOR_BUILD_FOLDER'] # only CI using docker
       begin


wpolicarpo:
@aharpervc I'm not sure how to write a proper test case for this. I can see there is a test `must run this test to prove we account for dropped connections` being skipped now and looks like it should be testing something similar to what we need here. Any suggestions?

aharpervc:
Yeah good question how can we force this behavior? What about something like this:
1. create a connection
2. run something like `kill @@spid` (not sure if this can be done all at once, might need to be queried, then templated, then executed)
3. with that same connection, attempt to run `select 1`
4. validate what happens when that fails

jeremyevans:
Even if this can't be cleanly tested, it's clearly better to raise an exception instead of returning false.  If you look at the README, it's clear that execute is never expected to return false, as none of the code that calls execute checks for false.  And apparently, none of the specs check for it returning false either.

bf4:
Just came across this from the rails sql server adapter source.  Is there a blocker to merging this? cc @metaskills

aharpervc:
> Just came across this from the rails sql server adapter source. Is there a blocker to merging this? cc @metaskills

There's no real hard blockers, however it would be great if we added a test for this. You can also contribute that test @bf4 if this is a priority for you.

bf4:
Should be easy for me to test given my local docker server keeps dying mid
request

aharpervc:
@wpolicarpo @bvogelzang was able to force various timeouts in https://github.com/rails-sqlserver/tiny_tds/pull/481, so maybe something can be done here that way? Alternately if there isn't enough energy to add a test, I would be okay merging regardless. I think the value is high enough on this one to do it anyway.

wpolicarpo:
@aharpervc I'm struggling to find time to look into this and some other adapter issues these days, but I can look into adding some tests, yes. I was already following that PR to get some ideas and implement them the same way here.

If you are planning on cutting a release anytime soon, we could merge this the way it is and I can open a new PR adding tests after if you're OK with that.

wpolicarpo:
@aharpervc I finally had time to look into this today. I think I will wait until #481 is merged and use the same approach to write some tests for this change. I could rebase my branch on top of that, but I would still need it to get merged first anyway. 

Do you have a rough idea of when that PR will be merged?

aharpervc:
> Do you have a rough idea of when that PR will be merged?

I expect we can merge it when the CI tests pass. I don't recall off the top of my head

aharpervc:
> I expect we can merge it when the CI tests pass. I don't recall off the top of my head what the blocker there was

Ben got the tests to pass, so we're moving forward. I included this PR in the 2.1.4-pre2 for validation: https://github.com/rails-sqlserver/tiny_tds/pull/486. I still think it'd be ideal to include a test in this PR if possible. Either way, can you add a bit to the "unreleased" section of the changelog @wpolicarpo?

wpolicarpo:
Thanks @aharpervc and @bvogelzang. I'll look into adding a test using toxiproxy this week.

aharpervc:
@wpolicarpo I saw you added a testing commit here https://github.com/rails-sqlserver/tiny_tds/pull/486/commits/1c2f56e5ff740c7a745bf566c455bacd6d1fbf1b, what do you think about adding it to this branch? Did Toxiproxy have the intended effect?

aharpervc:
@wpolicarpo I moved your commit that added a test from https://github.com/rails-sqlserver/tiny_tds/pull/486 to here and rebased on latest master (I realize this may be slightly disruptive since I force pushed, but it can be fixed if it's a problem)

bvogelzang:
@wpolicarpo I think changing the test so it uses the `timeout` toxic instead of the `down` toxic will get us what we want:

```ruby
client = new_connection timeout: 2, port: 1234
assert_client_works(client)
action = lambda { sleep(0.1); client.execute("select 1").do }

# Use toxiproxy to close the TCP socket after 10ms.
# We want TinyTds to fail to send the statement and raise an error immediately.
Toxiproxy[:sqlserver_test].toxic(:timeout, timeout: 10).apply do
    assert_match %r{failed to execute statement}i, e.message, 'ignore if non-english test run'
  end
end
```

wuarmin:
This PR would fix a bug in my web app, which would be very important for me.
Can it be merged soon?

wuarmin:
hey @bvogelzang 
is there a chance that this PR will be merged?

best regards

wpolicarpo:
I'm closing this PR because I'm unable to find time to look into it better and get it to a state that we can merge. If someone is able/wants to make it happen, please feel free to use the idea/code in here and open a new PR.

jeremyevans:
@wpolicarpo If you have time, can you clarify what about the PR made it unable to merge?  I'm not sure about the test, but the `client.c` change looks very straight forward, and even without a test, as long as it doesn't break any CI tests, seems like an improvement that should be merged.

Issue #445, opened 2019-10-09
Timeout does not work when tcp packets are dropped


bugg2844:
Scenario: Switching from a primary to a backup network, the firewall invalidates all existing tcp sessions and starts dropping packets.

Expected results: Configured timeout is hit and tiny_tds reports an error
Actual results: No response from tiny_tds until underlying tcp connection's retries expire (> 15 minutes on Ubuntu, for example: https://pracucci.com/linux-tcp-rto-min-max-and-tcp-retries2.html)

Simple sample code attached in timeout.zip file.  You'll need to simulate dropped packets in order to see the issue, which is shown about halfway down the attached freetds.log.  Freetds reports a timeout in the log, over and over and over, but tiny_tds does not return any error to the caller.  I am not sure if the issue is actually with tiny_tds or freetds.

Steps to reproduce:
1. Prereqs
    1. A firewall is required that can be configured to drop packets, in order to simulate the condition.  iptables is sufficient (e.g. apt-get install iptables), or pfctl on Mac.  
    2. Configure freetds to debug:
    	dump file = /tmp/freetds.log
	debug flags = 0xffff
2. Extract timeout.zip
3. Modify main.rb for your database connection
4. bundle install
5. bundle exec ruby main.rb
    1. You'll need to complete step 6 while main.rb is running.  Modify the loop as required.
6. Interrupt the database connection:
    1. For linux:
    iptables -I INPUT -s 1.2.3.4 -j DROP    (where 1.2.3.4 is your db's ip address)
    2. For mac:
        1.  Append to /etc/pf.conf (or create a custom pf.conf):
        block drop from 1.2.3.4 to any     (where 1.2.3.4 is your db's ip address)
        2. sudo pfctl /etc/pf.conf

At this point you should see the repeatedly ignored timeouts in freetds.log.  On Linux this will continue for over 15 minutes before the connection finally times out.  On Mac it takes around 30 seconds.  

You can restore your connection by replacing DROP with ACCEPT for iptables, or by removing the extra line from pf.conf and running pfctl again.

[freetds.log](https://github.com/rails-sqlserver/tiny_tds/files/3708515/freetds.log)
[timeout.zip](https://github.com/rails-sqlserver/tiny_tds/files/3708516/timeout.zip)

aharpervc:
[...]

Issue #78, opened 2012-04-05
multiple raised errors ... only last one captured


mikebarry:
I have an app that uses extensive stored procs and with a few tweaks, tiny_tds works great (thanks!). I do have one issue that if a proc has multiple raised errors (happens a lot when procs call procs), tiny_tds only captures the last error. I believe it's due to the
way tinytds_msg_handler calls rb_tinytds_raise_error which creates a new error instead of buffering them up.

I would love to submit a patch but I'm hitting the wall on my ruby internals knowledge (pointers welcome, I'm more than happy to tackle this if someone can point me in a halfway decent direction).

I do have this simple test case that displays the problem:

    require 'test_helper'

    class MultiErrorsTest < ActiveSupport::TestCase
  
      test 'should have multi errors' do

        @client = TinyTds::Client.new( 
          :dataserver => 'DEV', 
          :username => 'sa',
          :password => 'xxxxx',
          :tds_version => '42',
          :database => 'tempdb' 
        )

       begin
         @client.execute( 'DROP PROCEDURE inner_proc_test' ).do
         @client.execute( 'DROP PROCEDURE outer_proc_test' ).do
       rescue
         # do nothing ... we don't care about drop warnings
       end

       inner_proc_test  = @client.execute( "create procedure inner_proc_test as raiserror 20014 'Inner Proc Error'" ).do
       outer_proc_test  = @client.execute( "create procedure outer_proc_test as exec inner_proc_test raiserror 20014 'Outer Proc Error'" ).do

       begin 
         results = @client.execute( 'exec outer_proc_test' ).do
       rescue Exception => e
         assert_match( /inner/, e.message )
       end
    
      end
    end

This will fail because e is just the error from the outer_proc_test raise .. inner_proc_test raise is lost (and leaked?)

aharpervc:
I believe this will be addressed by https://github.com/rails-sqlserver/tiny_tds/pull/475

PR #475, opened 2020-07-24
Report stored procedure errors when using info messages

bvogelzang:
### Summary

Capturing errors while in a non-blocking state was originally structured to capture a single error. This was intentional in order to avoid capturing more generic info messages that FreeTDS might send before the Global VM Lock was obtained. In most circumstances this is what we want. However, now that we capture info messages it is possible that a info message will be stored and the actual runtime error will be mistakenly discarded as non-important. The result is that while a runtime error is reported in the database, a `TinyTds` error is never thrown and only the info message is handled. A subset of this problem is that only one info message can be captured while in non-blocking mode which prevents stored procedures from reporting multiple info messages to `TinyTds`.

To fix this issue, the reported messages are stored within a dynamic array of `tinytds_errordata` structs, then processed normally once the GVL is obtained.

Given the fact that we don't know the number of messages that will be sent, we dynamically manage and re-allocate memory for the `nonblocking_errors` array as needed. We can't use the ruby C API because it is not safe to call while in a non-blocking state as shown by #133.

### Example

Given this procedure:

```sql
CREATE PROCEDURE tinytds_TestPrintWithError
AS
    PRINT 'hello'
    RAISERROR('Error following print', 16, 1)
```

Result:

```ruby
@client = TinyTds::Client.new({host: 'localhost', username: 'tinytds', password: '', database: 'tinytdstest', message_handler: Proc.new { |m| puts "Info: #{m.message}" }})
@client.execute("exec tinytds_TestPrintWithError").do

# Before - no error raised

Info: hello
=> -1

# After - error properly raised

Info: hello
Traceback (most recent call last):
        ...
        1: from (irb):10:in `do'
TinyTds::Error (Error following print)

```

### Notes

- The memory management could use some extra attention. Curious to hear thoughts on how this should be handled or improved.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## (unreleased)
 
 * Improve handling of network related timeouts
+* Fix error reporting when preceded by info message
 
 ## 2.1.3
 
--- a/ext/tiny_tds/client.c
+++ b/ext/tiny_tds/client.c
@@ -24,25 +24,25 @@ VALUE opt_escape_regex, opt_escape_dblquote;
 
 // Lib Backend (Helpers)
 
-VALUE rb_tinytds_raise_error(DBPROCESS *dbproc, int is_message, int cancel, const char *error, const char *source, int severity, int dberr, int oserr) {
+VALUE rb_tinytds_raise_error(DBPROCESS *dbproc, tinytds_errordata error) {
   VALUE e;
   GET_CLIENT_USERDATA(dbproc);
-  if (cancel && !dbdead(dbproc) && userdata && !userdata->closed) {
+  if (error.cancel && !dbdead(dbproc) && userdata && !userdata->closed) {
     userdata->dbsqlok_sent = 1;
     dbsqlok(dbproc);
     userdata->dbcancel_sent = 1;
     dbcancel(dbproc);
   }
-  e = rb_exc_new2(cTinyTdsError, error);
-  rb_funcall(e, intern_source_eql, 1, rb_str_new2(source));
-  if (severity)
-    rb_funcall(e, intern_severity_eql, 1, INT2FIX(severity));
-  if (dberr)
-    rb_funcall(e, intern_db_error_number_eql, 1, INT2FIX(dberr));
-  if (oserr)
-    rb_funcall(e, intern_os_error_number_eql, 1, INT2FIX(oserr));
+  e = rb_exc_new2(cTinyTdsError, error.error);
+  rb_funcall(e, intern_source_eql, 1, rb_str_new2(error.source));
+  if (error.severity)
+    rb_funcall(e, intern_severity_eql, 1, INT2FIX(error.severity));
+  if (error.dberr)
+    rb_funcall(e, intern_db_error_number_eql, 1, INT2FIX(error.dberr));
+  if (error.oserr)
+    rb_funcall(e, intern_os_error_number_eql, 1, INT2FIX(error.oserr));
 
-  if (severity <= 10 && is_message) {
+  if (error.severity <= 10 && error.is_message) {
     VALUE message_handler = userdata && userdata->message_handler ? userdata->message_handler : Qnil;
     if (message_handler && message_handler != Qnil && rb_respond_to(message_handler, intern_call) != 0) {
       rb_funcall(message_handler, intern_call, 1, e);
@@ -57,6 +57,16 @@ VALUE rb_tinytds_raise_error(DBPROCESS *dbproc, int is_message, int cancel, cons
 
 
 // Lib Backend (Memory Management & Handlers)
+static void push_userdata_error(tinytds_client_userdata *userdata, tinytds_errordata error) {
+  // reallocate memory for the array as needed
+  if (userdata->nonblocking_errors_size == userdata->nonblocking_errors_length) {
+    userdata->nonblocking_errors_size *= 2;
+    userdata->nonblocking_errors = realloc(userdata->nonblocking_errors, userdata->nonblocking_errors_size * sizeof(tinytds_errordata));
+  }
+
+  userdata->nonblocking_errors[userdata->nonblocking_errors_length] = error;
+  userdata->nonblocking_errors_length++;
+}
 
 int tinytds_err_handler(DBPROCESS *dbproc, int severity, int dberr, int oserr, char *dberrstr, char *oserrstr) {
   static const char *source = "error";
@@ -105,6 +115,16 @@ int tinytds_err_handler(DBPROCESS *dbproc, int severity, int dberr, int oserr, c
       break;
   }
 
+  tinytds_errordata error_data = {
+    .is_message = 0,
+    .cancel = cancel,
+    .severity = severity,
+    .dberr = dberr,
+    .oserr = oserr
+  };
+  strncpy(error_data.error, dberrstr, ERROR_MSG_SIZE);
+  strncpy(error_data.source, source, ERROR_MSG_SIZE);
+
   /*
   When in non-blocking mode we need to store the exception data to throw it
   once the blocking call returns, otherwise we will segfault ruby since part
@@ -116,27 +136,9 @@ int tinytds_err_handler(DBPROCESS *dbproc, int severity, int dberr, int oserr, c
       dbcancel(dbproc);
       userdata->dbcancel_sent = 1;
     }
-
-    /*
-    If we've already captured an error message, don't overwrite it. This is
-    here because FreeTDS sends a generic "General SQL Server error" message
-    that will overwrite the real message. This is not normally a problem
-    because a ruby exception is normally thrown and we bail before the
-    generic message can be sent.
-    */
-    if (!userdata->nonblocking_error.is_set) {
-      userdata->nonblocking_error.is_message = 0;
-      userdata->nonblocking_error.cancel = cancel;
-      strncpy(userdata->nonblocking_error.error, dberrstr, ERROR_MSG_SIZE);
-      strncpy(userdata->nonblocking_error.source, source, ERROR_MSG_SIZE);
-      userdata->nonblocking_error.severity = severity;
-      userdata->nonblocking_error.dberr = dberr;
-      userdata->nonblocking_error.oserr = oserr;
-      userdata->nonblocking_error.is_set = 1;
-    }
-
+    push_userdata_error(userdata, error_data);
   } else {
-    rb_tinytds_raise_error(dbproc, 0, cancel, dberrstr, source, severity, dberr, oserr);
+    rb_tinytds_raise_error(dbproc, error_data);
   }
 
   return return_value;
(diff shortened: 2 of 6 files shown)


aharpervc:
This looks really good, and I also see that CI has passed, which is a good sign. I think this can probably be merged after 2.1.3 is released.

ext/tiny_tds/client.c
```diff
@@ -190,7 +196,6 @@ static void rb_tinytds_client_free(void *ptr) {
     cwrap->closed = 1;
     cwrap->userdata->closed = 1;
   }
-  xfree(cwrap->userdata);
```
bvogelzang, line 193: Just realized that this line change might have been left over from a rebase. I'll have to take another look to see if it's needed or not.

aharpervc:
Idea: can something be done to flush the error array, and if so, would that address this bug: https://github.com/rails-sqlserver/tiny_tds/issues/445

bvogelzang:
I can't think of a way to adjust this in a way that would ultimately solve the timeout issue. See my commit message here: https://github.com/rails-sqlserver/tiny_tds/compare/master...bvogelzang:network-timeout-handling. 

In cases of network failure I think the problem is:

1. `dbcancel` does not succeed because of the network failure
2. `dbsqlok` does not return due to above until the underlying network timeout on the os is hit. 
3. `TinyTds` does not throw an error in the expected timeout window.

To _really_ fix that issue we'd need some way to detect the difference between a timeout error on the happy path (with normal network) vs. the bad path (with network failture). Otherwise we can't branch accordingly. It would be nice if we could detect if `dbcancel` has failed for this purpose but it looks like it always returns `SUCCEED` at the moment 😞 https://github.com/FreeTDS/freetds/blob/master/src/dblib/dblib.c#L3411.

ext/tiny_tds/client.c
```diff
@@ -144,16 +145,21 @@ int tinytds_msg_handler(DBPROCESS *dbproc, DBINT msgno, int msgstate, int severi
...
-      userdata->nonblocking_error.cancel = is_message_an_error;
-      strncpy(userdata->nonblocking_error.error, msgtext, ERROR_MSG_SIZE);
-      strncpy(userdata->nonblocking_error.source, source, ERROR_MSG_SIZE);
-      userdata->nonblocking_error.severity = severity;
-      userdata->nonblocking_error.dberr = msgno;
-      userdata->nonblocking_error.oserr = msgstate;
-      userdata->nonblocking_error.is_set = 1;
-    }
+    /*
+    In the case of stored procedure execution we can receive multiple messages
```
aharpervc, line 149: This logic applies to all scenarios where we might get multiple messages, not just sp's, right?

ext/tiny_tds/client.c
```diff
@@ -57,6 +57,15 @@ VALUE rb_tinytds_raise_error(DBPROCESS *dbproc, int is_message, int cancel, cons
 
 
 // Lib Backend (Memory Management & Handlers)
+static void push_userdata_error(tinytds_client_userdata *userdata, tinytds_errordata error) {
+  // reallocate memory for the array as needed
+  if (userdata->nonblocking_errors_size == userdata->nonblocking_errors_length) {
+    userdata->nonblocking_errors_size *= 2;
+    userdata->nonblocking_errors = realloc(userdata->nonblocking_errors, userdata->nonblocking_errors_size * sizeof(tinytds_errordata));
+  }
```
aharpervc, line 65: This seems odd, so I have a few questions (not a pro with C, these are real questions)

1. When would `push_userdata_error` be called, and this if statement _not_ be true. I'd expect that every time we need to push an error on to the array, we'd need to resize the array.
2. Why is `nonblocking_errors_size` being doubled every time? Wouldn't it be preferable to incrementally grow the size in multiples of `sizeof(tinytds_errordata)`?
3. Why isn't the realloc like count x size, eg `realloc(userdata->nonblocking_errors, (userdata->nonblocking_errors_length + 1) * sizeof(tinytds_errordata))`?
3. Instead of tracking more state as a struct field, is anything gained by doing `sizeof(userdata->nonblocking_errors)` (assuming this would give the right answer, and if not why not?)

ext/tiny_tds/client.c
```diff
@@ -57,6 +57,15 @@ VALUE rb_tinytds_raise_error(DBPROCESS *dbproc, int is_message, int cancel, cons
...
+
+  userdata->nonblocking_errors[userdata->nonblocking_errors_length++] = error;
```
aharpervc, line 67: This method might get marginally simpler if `userdata->nonblocking_errors_length += 1` is added to the top of the method, then subsequent sizes and so forth are compared to that.

test/test_helper.rb
```diff
@@ -191,6 +199,13 @@ def sp_sql
         RETURN(420) |
     end
 
+    def sp_error_sql
+      %|CREATE PROCEDURE tinytds_TestPrintWithError
+        AS
+        PRINT 'hello'
+        RAISERROR('Error following print', 16, 1)|
```
aharpervc, line 206: This test is specific to "message then error". We might want another test which is "message followed by more messages".

Maybe like:

```ruby
def sp_several_prints_sql
  %|CREATE or alter PROCEDURE tinytds_TestSeveralPrints
    AS
    PRINT 'hello 1'
    PRINT 'hello 2'
    PRINT 'hello 3' |
end

# ---

it 'must raise an error for each of a series of `print` messages' do
  messages.clear
  action = lambda { @client.execute("EXEC tinytds_TestSeveralPrints").do }
    assert_equal ['hello 1', 'hello 2', 'hello 3'], messages.map { |e| e.message }, 'message list'
  end
end
```

ext/tiny_tds/client.h
```diff
@@ -5,9 +5,9 @@
 void init_tinytds_client();
 
 #define ERROR_MSG_SIZE 1024
+#define ERRORS_STACK_INIT_SIZE 2
```
aharpervc, line 8: What's special about 2?

aharpervc (changes requested):

ext/tiny_tds/client.c
bvogelzang: Yup, good catch. This was due to lack of understanding when I was first writing this up.

ext/tiny_tds/client.c
bvogelzang: If we attempt to push an error to the array and we don't have any more allocated space to add the new error we double the allocated size for the array. This allocates space for the new error as well as _future_ space that we may end up using. This is pretty standard practice for creating dynamic arrays. The reason being that reallocating the new space is expensive (`O(n)`) since the memory may have to be copied somewhere else in order to find the contiguous space for the new array size. As a result we want to reduce the number of times we have to do that operation.

1. `nonblocking_errors_size` represents the current _allocated_ size of the array. So these values are only ever the same if we've used up all of the currently allocated space.
2. The cost of memory allocation. See above.
3. The cost of memory allocation. See above.
4. `sizeof(userdata->nonblocking_errors)` would report the size of the pointer itself _not_ the size of the array that the pointer points to. So it wouldn't give the answer you're expecting.

ext/tiny_tds/client.h
bvogelzang: Nothing really. I picked a value that wasn't too greedy and that would support most use cases (0-2 errors reported) without needing to reallocate any memory. I'd be open to changing this to something else.

ext/tiny_tds/client.c
```diff
@@ -171,7 +177,7 @@ static void rb_tinytds_client_reset_userdata(tinytds_client_userdata *userdata)
   userdata->dbsqlok_sent = 0;
   userdata->dbcancel_sent = 0;
   userdata->nonblocking = 0;
-  userdata->nonblocking_error.is_set = 0;
+  userdata->nonblocking_errors_length = 0;
```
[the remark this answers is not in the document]
bvogelzang: We can't free the memory here if we're freeing the memory in `nogvl_cleanup`. When this is called there is nothing to free. I am open to adding `userdata->nonblocking_errors_size = 0` here for consistency or going the other route and removing the reset for `userdata->nonblocking_errors_length`.

I decided on initializing/freeing the array in `nogvl_setup` and `nogvl_cleanup` because that marks the start and end of the time in which we need to make use of the errors array. So the array is around just long enough for its purpose. If we want the errors array to have a longer life cycle I'd be open to discussing that. For example, now that we are capturing multiple errors we may want to add a way to access those errors after we raise the exception (see #78). In that case it might be worth keeping the array around longer.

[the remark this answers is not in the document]
aharpervc: >  I am open to adding userdata->nonblocking_errors_size = 0 here for consistency or going the other route and removing the reset for userdata->nonblocking_errors_length

I think the consistency would be preferable, as well as possibly a code comment as to why they're being zero'd but not free'd.

> I decided on initializing/freeing the array in nogvl_setup and nogvl_cleanup because that marks the start and end of the time in which we need to make use of the errors array. So the array is around just long enough for its purpose. 

Makes sense, thanks

ext/tiny_tds/client.c
aharpervc: Alright, thanks for explaining. I guess it makes sense to optimize for minimal number of allocation _operations_ rather than _total memory size_, as a strategy. I will defer to your judgement on that this is both the right strategy and the appropriate implementation of that strategy.

ext/tiny_tds/client.h
aharpervc: Makes sense

ext/tiny_tds/client.c
bvogelzang: I opted for incrementing at the end to prevent the need for `userdata->nonblocking_errors_length - 1` in other places. 

```suggestion
  userdata->nonblocking_errors[userdata->nonblocking_errors_length] = error;
  userdata->nonblocking_errors_length++;
```

bvogelzang:
@metaskills, @larskanis, @coderjoe, @wpolicarpo any chance for getting a review on this PR, https://github.com/rails-sqlserver/tiny_tds/pull/483, and https://github.com/rails-sqlserver/tiny_tds/pull/481

aharpervc (approved):

bvogelzang:
These changes are now available to test in [2.1.4.pre](https://rubygems.org/gems/tiny_tds)

aharpervc:
I'm testing this with `2.1.4.pre` and I'm still not seeing both message from your example sp. I get the initial `hello`, but then no subsequent error. If I decrease the severity to a message, I still only see `hello`.

aharpervc:
While testing https://github.com/rails-sqlserver/tiny_tds/pull/481 I noticed that only `raiserror ... with nowait` seems to be displayed when there's a timeout, while `print` and `raiserror` without nowait didn't. 

- is this expected?
- can we figure out how to flush whatever buffer has these messages when there's an error?
- should we have tests for this situation?

example script (basically the same as the other PR):

```ruby
require 'tiny_tds'

def w(*args)
  puts [Thread.current.object_id, *args].join ' '
end

t = Thread.new do
  i = 0
  loop do
    i += 1
    w i
    sleep(1)
  end
end

c = TinyTds::Client.new({
  host: 'localhost',
  username: 'sa',
  password: 'Testing123@@',

  timeout: 5,
  login_timeout: 5,

  message_handler: ->(e) { w e },
})

w TinyTds::VERSION

sql = <<~SQL
create or alter procedure tinytds_TestPrintWithError
as
begin
    print 'print 1' -- nope
    raiserror('raiserror 1', 1, 1) -- nope
    -- raiserror('raiserror 2', 1, 1) with nowait -- yep, and uncommenting this will show all 3 messages

    waitfor delay '00:01:00'
end
SQL

w c.execute(sql).to_a
w c.execute('exec tinytds_TestPrintWithError').to_a

t.exit
```

`raiserror with nowait` commented out: 

```
$ bundle exec ruby .\scratch.rb
24279960 1
18084080 2.1.4.pre
18084080
24279960 2
24279960 3
24279960 4
24279960 5
24279960 6
Traceback (most recent call last):
        2: from ./scratch.rb:42:in `<main>'
        1: from ./scratch.rb:42:in `to_a'
./scratch.rb:42:in `each': Adaptive Server connection timed out (TinyTds::Error)
```

uncommented:

```
$ bundle exec ruby .\scratch.rb
28287960 1
18149600 2.1.4.pre
18149600
28287960 2
28287960 3
28287960 4
28287960 5
28287960 6
18149600 print 1
18149600 raiserror 1
18149600 raiserror 2
Traceback (most recent call last):
        2: from ./scratch.rb:42:in `<main>'
        1: from ./scratch.rb:42:in `to_a'
./scratch.rb:42:in `each': Adaptive Server connection timed out (TinyTds::Error)
```

bvogelzang:
> is this expected?

I'm not exactly sure. This really just comes down to the order of _when_ freetds notifies tiny_tds of the info message. With `nowait` in this scenario freetds calls the dbmsghandle before the dberrhandle. Without it, freetds calls the dbmsghandle after calling the dberrhandle. It's unclear to me if this is what _should_ be happening or if this is a ordering bug within freetds.

> can we figure out how to flush whatever buffer has these messages when there's an error?

Yes, we can definitely do this. These info messages are being pushed to our nonblocking_errors array which means we have access to them in `novgl_cleanup`. We just need to look past any errors in the array for any remaining info messages. e.g

```c
for (short int i = 0; i < userdata->nonblocking_errors_length; i++) {
    tinytds_errordata error = userdata->nonblocking_errors[i];

    // look ahead to drain any info messages ahead of raising an error
    if (!error.is_message) {
      for (short int j = i; j < userdata->nonblocking_errors_length; j++) {
        tinytds_errordata msg_error = userdata->nonblocking_errors[j];
        if (msg_error.is_message) {
          rb_tinytds_raise_error(client,
            msg_error.is_message,
            msg_error.cancel,
            msg_error.error,
            msg_error.source,
            msg_error.severity,
            msg_error.dberr,
            msg_error.oserr
          );
        }
      }
    }

    rb_tinytds_raise_error(client,
      error.is_message,
      error.cancel,
      error.error,
      error.source,
      error.severity,
      error.dberr,
      error.oserr
    );
  }
```
I'm trying to come up with a reason we _wouldn't_ want to do this but can't really think of a downside.

> should we have tests for this situation?

Yes, we probably should.

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## (unreleased)
 
+* Fix error reporting when preceeded by info message
```
aharpervc, line 3: "preceded"

aharpervc:
These changes look good to me. We've been testing 2.1.4-pre2 in production for a week or so and this change seems to have the intended outcome.

Can you squash/edit your commits down to whatever's appropriate for master history before merging?

PR #481, opened 2020-10-05
Improve handling of network related timeouts

### LINKED ISSUE #445: Timeout does not work when tcp packets are dropped

bvogelzang:
### Summary

In cases of network failure with a command batch running - configured timeouts were not being respected. This meant that the time elapsed before a `TinyTds::Error` was thrown would extend well beyond the configured timeout #445.

The reason for this is due to the fact that `dbcancel` does not work under these conditions. The `tinytds_err_handler` is called when the timeout is reached, we return `INT_TIMEOUT` and attempt to cancel the command batch using `dbcancel`. That call fails which means we continue to wait for the command batch to return even when subsequent timeout errors are reported.

Ideally to fix this we would detect if `dbcancel` fails and return `INT_CANCEL` in these circumstances. Unfortunately, we can't do this given that `dbcancel` [will always return `SUCCEED`](https://github.com/FreeTDS/freetds/blob/master/src/dblib/dblib.c#L3389-L3412)

As an alternative, when a timeout occurs, we set a flag and use that to return `INT_CANCEL` on the next timeout error we encounter. This allows the pending command batch to return and `nogvl_cleanup` can raise the ruby exception. This works well except for the fact that it will take two timeout periods to elapse before sending `INT_CANCEL`. One reported timeout error to set the flag and another one to act on it. To get around this we can use `dbsetinterrupt` to check the flag periodically while waiting on a read from the server, return `INT_CANCEL` sooner, and raise the timeout exception.

This shouldn't have any affect on normal timeout conditions due to the fact that `dbcancel` will actually succeed and cause the normal flow before the interrupt can be called/handled. This is good because in these situations we still want the db process to remain in working condition.

Fixes #445.

### Notes

- I can't come up with a good way to write a test for this scenario. I would love suggestions on how we might approach this.
- Returning `INT_CANCEL` means that the db process will be killed. It's not ideal that after some timeout errors the connection will still be usable, while for others it will not. This is how things behaved prior to this PR but wanted to call it out. Is this worth adding as a note in the readme?
- In my testing, the interrupt was called once per second. This means the timeout exception will be raised 1 second after hitting the configured timeout. I have no idea what the interrupt interval is in other environments so it is possible that this could vary.
- Other solutions I considered were trying to kill the thread spawned by `rb_thread_call_without_gvl` and raising the error directly from the error handler using `rb_thread_call_with_gvl` but I couldn't get either solution working due to the fact that I couldn't differentiate between timeouts caused by normal db processing and those caused by network related issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ Creating a new client takes a hash of options. For valid iconv encoding options,
 * :appname - Short string seen in SQL Servers process/activity window.
 * :tds_version - TDS version. Defaults to "7.3".
 * :login_timeout - Seconds to wait for login. Default to 60 seconds.
-* :timeout - Seconds to wait for a response to a SQL command. Default 5 seconds. Prior to 1.0rc5, FreeTDS was unable to set the timeout on a per-client basis, permitting only a global timeout value. This means that if you're using an older version, the timeout values for all clients will be overwritten each time you instantiate a new `TinyTds::Client` object. If you are using 1.0rc5 or later, all clients will have an independent timeout setting as you'd expect.
+* :timeout - Seconds to wait for a response to a SQL command. Default 5 seconds. Prior to 1.0rc5, FreeTDS was unable to set the timeout on a per-client basis, permitting only a global timeout value. This means that if you're using an older version, the timeout values for all clients will be overwritten each time you instantiate a new `TinyTds::Client` object. If you are using 1.0rc5 or later, all clients will have an independent timeout setting as you'd expect. Timeouts caused by network failure will raise a timeout error 1 second after the configured timeout limit is hit (see [#481](https://github.com/rails-sqlserver/tiny_tds/pull/481) for details).
 * :encoding - Any valid iconv value like CP1251 or ISO-8859-1. Default UTF-8.
 * :azure - Pass true to signal that you are connecting to azure.
 * :contained - Pass true to signal that you are connecting with a contained database user.
@@ -321,6 +321,10 @@ By default row caching is turned on because the SQL Server adapter for ActiveRec
 
 TinyTDS takes an opinionated stance on how we handle encoding errors. First, we treat errors differently on reads vs. writes. Our opinion is that if you are reading bad data due to your client's encoding option, you would rather just find `?` marks in your strings vs being blocked with exceptions. This is how things wold work via ODBC or SMS. On the other hand, writes will raise an exception. In this case we raise the SYBEICONVO/2402 error message which has a description of `Error converting characters into server's character set. Some character(s) could not be converted.`. Even though the severity of this message is only a `4` and TinyTDS will automatically strip/ignore unknown characters, we feel you should know that you are inserting bad encodings. In this way, a transaction can be rolled back, etc. Remember, any database write that has bad characters due to the client encoding will still be written to the database, but it is up to you rollback said write if needed. Most ORMs like ActiveRecord handle this scenario just fine.
 
+
+## Timeout Error Handling
+
+TinyTDS will raise a `TinyTDS::Error` when a timeout is reached based on the options supplied to the client. Depending on the reason for the timeout, the connection could be dead or alive. When db processing is the cause for the timeout, the connection should still be usable after the error is raised. When network failure is the cause of the timeout, the connection will be dead. If you attempt to execute another command batch on a dead connection you will see a `DBPROCESS is dead or not enabled` error. Therefore, it is recommended to check for a `dead?` connection before trying to execute another command batch.
 
 ## Binstubs
 
--- a/README.md
+++ b/README.md
@@ -419,17 +419,20 @@ First, clone the repo using the command line or your Git GUI of choice.
 $ git clone git@github.com:rails-sqlserver/tiny_tds.git
 ```
 
-After that, the quickest way to get setup for development is to use [Docker](https://www.docker.com/). Assuming you have [downloaded docker](https://www.docker.com/products/docker) for your platform and you have , you can run our test setup script.
+After that, the quickest way to get setup for development is to use [Docker](https://www.docker.com/). Assuming you have [downloaded docker](https://www.docker.com/products/docker) for your platform, you can use [docker-compose](https://docs.docker.com/compose/install/) to run the necessary containers for testing.
 
 ```shell
-$ ./test/bin/setup.sh
+$ docker-compose up -d
 ```
 
-This will download our SQL Server for Linux Docker image based from [microsoft/mssql-server-linux/](https://hub.docker.com/r/microsoft/mssql-server-linux/). Our image already has the `[tinytdstest]` DB and `tinytds` users created. Basically, it does the following.
+This will download our SQL Server for Linux Docker image based from [microsoft/mssql-server-linux/](https://hub.docker.com/r/microsoft/mssql-server-linux/). Our image already has the `[tinytdstest]` DB and `tinytds` users created. This will also download a [toxiproxy](https://github.com/shopify/toxiproxy) Docker image which we can use to simulate network failures for tests. Basically, it does the following.
 
 ```shell
+$ docker network create main-network
 $ docker pull metaskills/mssql-server-linux-tinytds
-$ docker run -p 1433:1433 -d metaskills/mssql-server-linux-tinytds
+$ docker run -p 1433:1433 -d --name sqlserver --network main-network metaskills/mssql-server-linux-tinytds
+$ docker pull shopify/toxiproxy
+$ docker run -p 8474:8474 -p 1234:1234 -d --name toxiproxy --network main-network shopify/toxiproxy
 ```
 
 If you are using your own database. Make sure to run these SQL commands as SA to get the test database and user installed.
--- a/ext/tiny_tds/client.c
+++ b/ext/tiny_tds/client.c
@@ -86,7 +86,13 @@ int tinytds_err_handler(DBPROCESS *dbproc, int severity, int dberr, int oserr, c
       but we don't ever want to automatically retry. Instead have the app
       decide what to do.
       */
-      return_value = INT_TIMEOUT;
+      if (userdata->timing_out) {
+        return INT_CANCEL;
+      }
+      else {
+        userdata->timing_out = 1;
+        return_value = INT_TIMEOUT;
+      }
       cancel = 1;
       break;
 
@@ -163,6 +169,33 @@ int tinytds_msg_handler(DBPROCESS *dbproc, DBINT msgno, int msgstate, int severi
     rb_tinytds_raise_error(dbproc, !is_message_an_error, is_message_an_error, msgtext, source, severity, msgno, msgstate);
   }
   return 0;
+}
+
+/*
+Used by dbsetinterrupt -
+This gets called periodically while waiting on a read from the server
+Right now, we only care about cases where a read from the server is
+taking longer than the specified timeout and dbcancel is not working.
+In these cases we decide that we actually want to handle the interrupt
+*/
+static int check_interrupt(void *ptr) {
+  GET_CLIENT_USERDATA((DBPROCESS *)ptr);
+  return userdata->timing_out;
+}
+
+/*
+Used by dbsetinterrupt -
+This gets called if check_interrupt returns TRUE.
+Right now, this is only used in cases where a read from the server is
+taking longer than the specified timeout and dbcancel is not working.
+Return INT_CANCEL to abort the current command batch.
+*/
+static int handle_interrupt(void *ptr) {
+  GET_CLIENT_USERDATA((DBPROCESS *)ptr);
+  if (userdata->timing_out) {
+    return INT_CANCEL;
+  }
+  return INT_CONTINUE;
 }
 
 static void rb_tinytds_client_reset_userdata(tinytds_client_userdata *userdata) {
@@ -381,6 +414,7 @@ static VALUE rb_tinytds_connect(VALUE self, VALUE opts) {
       }
     }
     dbsetuserdata(cwrap->client, (BYTE*)cwrap->userdata);
+    dbsetinterrupt(cwrap->client, check_interrupt, handle_interrupt);
     cwrap->userdata->closed = 0;
     if (!NIL_P(database) && (azure != Qtrue)) {
       dbuse(cwrap->client, StringValueCStr(database));
--- a/test/client_test.rb
+++ b/test/client_test.rb
@@ -68,6 +68,9 @@ class ClientTest < TinyTds::TestCase
   end
 
   describe 'With in-valid options' do
+    before(:all) do
+      init_toxiproxy
+    end
 
     it 'raises an argument error when no :host given and :dataserver is blank' do
       assert_raises(ArgumentError) { new_connection :dataserver => nil, :host => nil }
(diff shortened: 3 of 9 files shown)


ext/tiny_tds/client.c
```diff
@@ -165,6 +171,29 @@ int tinytds_msg_handler(DBPROCESS *dbproc, DBINT msgno, int msgstate, int severi
...
+
+/*
+Used by dbsetinterrupt -
+This gets called if check_interrupt returns TRUE.
+Right now, this is only used in cases where a read from the server is
+taking longer than the specified timeout and dbcancel is not working.
+Return INT_CANCEL to abort the current command batch.
+*/
+static int handle_interrupt(void *ptr) {
+  return INT_CANCEL;
```
aharpervc, line 194: 1. Is it possible for `userdata->timing_out` to go false _after_ `check_interrupt` but _before_ `handle_interrupt`? I wonder if we should do something here like `if (userdata->timing_out) then return INT_CANCEL` to avoid that possibility, and then also maybe set up a pattern for any future interrupt handler logic.
2. Who is responsible for ultimately closing the connection if this is a permanent network failure? Eg, what is the outcome of `#dead?` and `#closed?` after this handler returns? https://github.com/rails-sqlserver/tiny_tds/blob/d8cb94e1033d124dfdd5a08780a530fd8c995101/ext/tiny_tds/client.c#L227-L235

ext/tiny_tds/client.c
```diff
@@ -392,6 +421,7 @@ static VALUE rb_tinytds_connect(VALUE self, VALUE opts) {
     } else {
       cwrap->identity_insert_sql = "SELECT CAST(SCOPE_IDENTITY() AS bigint) AS Ident";
     }
+    dbsetinterrupt(cwrap->client, check_interrupt, handle_interrupt);
```
aharpervc, line 424: Is it possible for a connection to time out between `dbopen` above and this line? I wonder if this is like the first thing we should set up, before trying to send any db commands like `dbuse`. Maybe right after `dbsetuserdata`?

aharpervc:
This looks like a great contribution, very nice.

> * I can't come up with a good way to write a test for this scenario. I would love suggestions on how we might approach this.

How did you test it during development?

> * Returning `INT_CANCEL` means that the db process will be killed. It's not ideal that after some timeout errors the connection will still be usable, while for others it will not. This is how things behaved prior to this PR but wanted to call it out. Is this worth adding as a note in the readme?

It seems like a really good idea to write this down somewhere. I am also a little unclear when a timeout will cause a connection to die and when it won't.

> * In my testing, the interrupt was called once per second. This means the timeout exception will be raised 1 second after hitting the configured timeout. I have no idea what the interrupt interval is in other environments so it is possible that this could vary.

Do you know who's responsible for defining that one second period? Is that tiny tds? freetds? someone else?

> * Other solutions I considered were trying to kill the thread spawned by `rb_thread_call_without_gvl` and raising the error directly from the error handler using `rb_thread_call_with_gvl` but I couldn't get either solution working due to the fact that I couldn't differentiate between timeouts caused by normal db processing and those caused by network related issues.

I think your current solution seems reasonable. What do you mean "normal db processing" though?

ext/tiny_tds/client.c
bvogelzang: 1. Great point. This seems like a possibility (though very slim) and I love your suggestion. I'll go ahead and add that check for safety.
2. I'll check up on that and report back. I had assumed both would return `false`.

bvogelzang:
> How did you test it during development?

I had a scratch script that would use `pfctl` to drop the connection.

```ruby
IO.write('/etc/pf.conf', File.open('/etc/pf.conf') { |f| f.read.gsub(/#block drop/, "block drop") })
system('pfctl -f /etc/pf.conf > /dev/null 2>&1')
```

Something like that isn't going to be cross platform and required a little bit of manual setup. If we can come up with a reliable set of commands to drop/restore a network connection across platforms then we can likely write a test for this case. I doubt it's desirable to drop network across the whole system just to run a test though so it would ideally work for just the test host.

> It seems like a really good idea to write this down somewhere. I am also a little unclear when a timeout will cause a connection to die and when it won't.

I'll take an initial stab at drafting something for the readme

> Do you know who's responsible for defining that one second period? Is that tiny tds? freetds? someone else?

Initially I couldn't find it but it looks like freetds defines this at 1 second: https://github.com/FreeTDS/freetds/blob/cfa161498673d91a963d173561cbc99835da1cfa/src/tds/net.c#L622. Which means it should be consistent.

> What do you mean "normal db processing" though?

I mean in cases where the timeout is triggered while waiting for results from the db and a network connection is up. e.g. the timeout is configured for 1 second but sql `waitfor delay 00:00:02` causes a timeout.

ext/tiny_tds/client.c
bvogelzang: Tested for both this PR and `2.1.3`. Both had the same result:

Timeouts with a working connection:
```
Client dead?: false
Client closed?: false
```

Network timeouts:
```
Client dead?: true
Client closed?: false
```

So it is currently the callers responsibility to close the connection.

aharpervc:
>  If we can come up with a reliable set of commands to drop/restore a network connection across platforms then we can likely write a test for this case.

SQL Server is running in a container, right? Does pausing the container after creating a connection have the same effect? Or maybe we can do something like this to pause the docker network temporarily: https://stackoverflow.com/a/56218103

We can also conditionally skip the test based on env vars, etc. I think it would be nice to have _something_ even if it's silly and shells out to docker.

bvogelzang:
Great suggestion. Pausing the container _does_ have a similar effect. I've gone ahead and refactored an existing test we were previously skipping altogether. If you have a better suggestion for the skip logic I'd be happy to incorporate it.

README.md
```diff
@@ -110,7 +110,7 @@ Creating a new client takes a hash of options. For valid iconv encoding options,
...
 * :tds_version - TDS version. Defaults to "7.3".
 * :login_timeout - Seconds to wait for login. Default to 60 seconds.
-* :timeout - Seconds to wait for a response to a SQL command. Default 5 seconds. Prior to 1.0rc5, FreeTDS was unable to set the timeout on a per-client basis, permitting only a global timeout value. This means that if you're using an older version, the timeout values for all clients will be overwritten each time you instantiate a new `TinyTds::Client` object. If you are using 1.0rc5 or later, all clients will have an independent timeout setting as you'd expect.
+* :timeout - Seconds to wait for a response to a SQL command. Default 5 seconds. Prior to 1.0rc5, FreeTDS was unable to set the timeout on a per-client basis, permitting only a global timeout value. This means that if you're using an older version, the timeout values for all clients will be overwritten each time you instantiate a new `TinyTds::Client` object. If you are using 1.0rc5 or later, all clients will have an independent timeout setting as you'd expect. Timeouts caused by network failure will raise a timeout error 1 second after the configured timeout limit is hit.
```
aharpervc, line 113: "(see PR #481 for details)"?

test/client_test.rb
```diff
@@ -129,27 +129,24 @@ class ClientTest < TinyTds::TestCase
...
+        docker_container('pause')
         action = lambda { client.execute('SELECT 1 as [one]').each }
         assert_raise_tinytds_error(action) do |e|
           assert_equal 20003, e.db_error_number
           assert_equal 6, e.severity
           assert_match %r{timed out}i, e.message, 'ignore if non-english test run'
         end
       ensure
-        STDOUT.puts "Reconnect network!"
-        sleep 10
```
aharpervc, line 148: Keeping at least a little sleep might be a good idea, I don't know if we can rely on SQL Server coming online instantly when the container is unpaused. I actually put in a sleep after each docker command just to be sure. What about an optional keyword parameter to the `docker_container` method?

test/client_test.rb
```diff
@@ -129,27 +129,24 @@ class ClientTest < TinyTds::TestCase
...
+        docker_container('unpause')
         action = lambda { client.execute('SELECT 1 as [one]').each }
         assert_raise_tinytds_error(action) do |e|
           assert_equal 20047, e.db_error_number
-          assert_equal 1, e.severity
+          assert_equal 9, e.severity
```
aharpervc, line 149: Interesting that this is changed, because for me the test is currently failing because the error severity actually is `1`. Now I'm surprised/worried if you and I are getting different severity levels for the same error.

test/client_test.rb
bvogelzang: Good point. Are you suggesting this `docker_container('sleep', wait_for: 1)`?

aharpervc: Yep

test/client_test.rb
bvogelzang: I wonder if this has to do with different versions of freetds? I'm currently running `1.2.3` on my machine. Also, it's worth noting that this test on Travis CI is passing which looks to be running `1.1.24`. If you try out one of those versions does the severity change?

aharpervc: Hm, that is definitely a (sad) possibility. I'm running v1.00.112. I'll try and upgrade and see if that changes things.

test/test_helper.rb
```diff
@@ -44,7 +44,7 @@ def sqlserver?
       self.class.sqlserver?
     end
 
-    def close_client(client=@client)
+    def close_client(client=(@client ||= nil))
```
aharpervc, line 47: This seems really odd. Is this doing an assignment to the `@client` instance variable? When when the method is called, I assume? What's the scenario where this is happening and we want `@client` to be nil?

bvogelzang: I agree it's odd. It's strange that we're even calling this method without a param in the first place when `@client` has not been initialized. This was an attempt to get rid of `warning: instance variable @client not initialized` warnings while running tests locally. I don't feel strongly about keeping it since we probably should just fix the places we're calling this from.

aharpervc: Would this do it?

```ruby
def close_client(client=nil)
  client ||= @client
  (...existing code...)
end
```

bvogelzang: Yeah I think so. I'll give it a try.

bvogelzang: Well, now I'm really confused because that change causes some tests to fail and I have no idea why 🤷

aharpervc: Another option would be to revert all this just for this PR, and put in a future PR to fix the warning

test/client_test.rb
bvogelzang: This commit would indicate it is in fact the underlying freetds version: https://github.com/FreeTDS/freetds/commit/66423608f405d5714b047a67652d8a6015094f29. Not sure what to do here...maybe a comment in the test explaining that the test could fail due to severity but that's ok?

aharpervc: We have a couple options

1. just change it (preferably in a commit by itself with a clear explanation & link over to that freetds commit)
2. do 1, plus changing the assertion to `assert_includes [1,9], e.severity`
3. do 1, but also bump the minimum supported version of FreeTDS

I vote option 2 because it's maximally clear about what could happen for various TinyTDS users. I don't think this particular situation warrants option 3.

bvogelzang: I agree option 2 is the best here. I'll update. Thanks!

aharpervc (approved):

bvogelzang:
These changes are now available to test in [2.1.4.pre](https://rubygems.org/gems/tiny_tds)

aharpervc:
> These changes are now available to test in [2.1.4.pre](https://rubygems.org/gems/tiny_tds)

This is probably fine for what it is, however we still have some timeout related problems. For example, when I run this scratch script and kill sql server, the main thread appears to hang forever instead of raising a timeout error.

1. run a sql server container
2. run this scratch script
3. immediately stop the sql server container
4. counter thread runs forever, main thread hangs instead of timeout

```ruby
require 'tiny_tds'

def w(*args)
  puts [Thread.current.object_id, *args].join ' '
end

t = Thread.new do
  i = 0
  loop do
    i += 1
    w i
    sleep(1)
  end
end

c = TinyTds::Client.new({
  host: 'localhost',
  username: 'sa',
  password: 'Testing123@@',

  timeout: 5,
  login_timeout: 5,

  message_handler: ->(e) { w e },
})

w TinyTds::VERSION

sql = <<~SQL
create or alter procedure tinytds_TestPrintWithError
as
begin
    raiserror('raiserror', 1, 1) with nowait
    waitfor delay '00:01:00'
end
SQL

w c.execute(sql).to_a
w c.execute('exec tinytds_TestPrintWithError').to_a

t.exit
```

```
$ bundle exec ruby .\scratch.rb
23804380 1
4649200 2.1.4.pre
4649200
23804380 2
23804380 3
23804380 4
23804380 5
23804380 6
23804380 7
23804380 8
23804380 9
23804380 10
23804380 11
23804380 12
23804380 13
23804380 14
23804380 15
(...snip...)
```

I'm not sure if this particular problem should be addressed by this PR, or we should just remember it's still a problem and tackle it separately.

sixfeetover:
One suggestion for how to test this is to use [ToxiProxy](https://github.com/shopify/toxiproxy). It's a tool for simulating different kinds of network failure. You can boot toxiproxy container, connect to mssql through it, and then kill the connection, add jitter or latency, etc.

bvogelzang:
I was unable to reproduce using your scratch script. I used [ToxiProxy](https://github.com/shopify/toxiproxy) to test network failures (slow close, timeout, and down) as well as manually pausing/stopping my sqlserver docker container. In all instances I hit a timeout at the expected point (at 6 seconds).

test/client_test.rb
```diff
@@ -129,30 +132,40 @@ class ClientTest < TinyTds::TestCase
...
-        STDOUT.puts "Disconnect network!"
-        sleep 10
-        STDOUT.puts "This should not get stuck past 6 seconds!"
-        action = lambda { client.execute('SELECT 1 as [one]').each }
-        assert_raise_tinytds_error(action) do |e|
-          assert_equal 20003, e.db_error_number
-          assert_equal 6, e.severity
-          assert_match %r{timed out}i, e.message, 'ignore if non-english test run'
+        action = lambda { client.execute("waitfor delay '00:00:02'").do }
+        # Delay the TCP socket from closing until delay has elapsed.
```
aharpervc, line 141: "Delay the TCP socket from closing until delay has elapsed." <-- _which_ delay? There's several here and it isn't 100% obvious which parts are important to us.

- 1s = tinytds execution timeout
- 2s = duration of sql statement
- 0.5s = delay for something? Is this killing the network connection after 500ms?

Is there any benefit to either making these whole integers (1,2,3) and/or naming them for clarity?

aharpervc (changes requested):

test/client_test.rb
bvogelzang: I went ahead and added clarifying comments about what these tests are trying to achieve and why. Let me know if I've made it clear enough.

aharpervc: It makes more sense now, thanks

aharpervc:
These changes look good to me. We've been testing 2.1.4-pre2 in production for a week or so without any ill effect from this PR.

Can you squash/edit your commits down to whatever's appropriate for master history before merging?

PR #494, opened 2021-05-04
Fix Appveyor builds by refreshing the package list

bvogelzang:
Appveyor builds were breaking due to an error related to pacman when running `bundle install`:

```
warning: database file for 'ucrt64' does not exist (use '-Sy' to download)
warning: database file for 'clang64' does not exist (use '-Sy' to download)
error: failed to prepare transaction (could not find database)
```

Running `pacman --noconfirm -Sy` to update the package list resolves this error and the builds pass successfully.


aharpervc (approved):
🤦‍♂️

PR #496, opened 2021-05-11
Fix for loop initial declarations C99 error

### LINKED ISSUE #495: 2.1.4 Compilation error on Amazon Linux 1  loop initial declarations are only allowed in C99 mode
## Environment

**Operating System**

Amazon Linux 1 (`4.14.225-121.362.amzn1.x86_64`) with GCC 4.8 (`x86_64-amazon-linux gcc version 4.8.5 20150623 (Red Hat 4.8.5-28) (GCC)`)

Problem occurs only on tiny_tds v 2.1.4.

```
# tsql -C
Compile-time settings (established with the "configure" script)
                            Version: freetds v1.00.112
             freetds.conf directory: /usr/local/etc
     MS db-lib source compatibility: no
        Sybase binary compatibility: no
                      Thread safety: yes
                      iconv library: yes
                        TDS version: 7.4
                              iODBC: no
                           unixodbc: no
              SSPI "trusted" logins: no
                           Kerberos: no
                            OpenSSL: yes
                             GnuTLS: no
                               MARS: no
```

**FreeTDS Version**

Any, tested on FreeTDS 1.2.21 , 1.1.40, 1.0.86

## Description

```
result.c:107:7: error: ‘for’ loop initial declarations are only allowed in C99 mode
       for (short int j = i; j < userdata->nonblocking_errors_length; j++) {
```

bvogelzang:
fixes #495

### DIFF
--- a/ext/tiny_tds/result.c
+++ b/ext/tiny_tds/result.c
@@ -99,12 +99,14 @@ static void nogvl_cleanup(DBPROCESS *client) {
   Now that the blocking operation is done, we can finally throw any
   exceptions based on errors from SQL Server.
   */
-  for (short int i = 0; i < userdata->nonblocking_errors_length; i++) {
+  short int i;
+  for (i = 0; i < userdata->nonblocking_errors_length; i++) {
     tinytds_errordata error = userdata->nonblocking_errors[i];
 
     // lookahead to drain any info messages ahead of raising error
     if (!error.is_message) {
-      for (short int j = i; j < userdata->nonblocking_errors_length; j++) {
+      short int j;
+      for (j = i; j < userdata->nonblocking_errors_length; j++) {
         tinytds_errordata msg_error = userdata->nonblocking_errors[j];
         if (msg_error.is_message) {
           rb_tinytds_raise_error(client, msg_error);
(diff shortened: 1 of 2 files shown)


ext/tiny_tds/result.c
```diff
@@ -99,12 +99,14 @@ static void nogvl_cleanup(DBPROCESS *client) {
...
-  for (short int i = 0; i < userdata->nonblocking_errors_length; i++) {
+  short int i;
+  short int j;
+  for (i = 0; i < userdata->nonblocking_errors_length; i++) {
     tinytds_errordata error = userdata->nonblocking_errors[i];
 
     // lookahead to drain any info messages ahead of raising error
     if (!error.is_message) {
-      for (short int j = i; j < userdata->nonblocking_errors_length; j++) {
+      for (j = i; j < userdata->nonblocking_errors_length; j++) {
```
aharpervc, line 109: Is the idea here to declare `j` outside of the `i` loop so as to only allocate once instead of allocating `i` times?

aharpervc (approved):
The code looks good to me (I guess, not a C expert). Can you also confirm this works with AL1? Should we cut another prerelease and have @januszm verify?

januszm (approved):

ext/tiny_tds/result.c
bvogelzang: Actually that wasn't my intention. I should probably move the `j` declaration inside the `i` loop.

PR #514, opened 2022-03-31
[Experimental] CircleCI

bryanwieg:
Extention of work by [aharpervc](https://github.com/aharpervc) here: https://github.com/veracross/tiny_tds/pull/1

Results here: https://app.circleci.com/pipelines/github/bryanwieg/tiny_tds?branch=circleci&filter=all


aharpervc:
Great progress, this is huge. Next steps: 

1. restore the ruby version parameterization (I'm thinking `major.minor` if possible, and let the patch version float. Not sure if this will be possible if we need the FULL version to template a url)
2. restore the linux job(s)
3. make sure we're able to use all reasonable caching to make things as speedy as possible

That'd be enough for this PR. However I'd like to do more at some point, but it could be a second PR (by you or someone else if you don't have time)

1. investigate if we can make a macOS job
2. set up integration with CircleCI's test reporting: https://circleci.com/docs/2.0/collect-test-data/
4. set up integration with CircleCI's artifacts collection: https://circleci.com/docs/2.0/artifacts/ (this way, people observing PR's can grab release gem instead of building from source)

Also we should consider restructuring the jobs workflow(s). For example, we might benefit from having "test" be separate from "build", and then require test to succeed before build runs. That probably isn't critical though.

@bvogelzang any thoughts?

bryanwieg:
This looks good, first three points look good. I need to clean things up.

I was wondering about the tests. In appveyor, they have a section spinning up SQL Server and executing some rake tests. I suppose there are also tests for Linux. Do these need to be implemented in circleci as well?

I unfortunately do not have a mac, so I cannot help with that.

bvogelzang:
Wow. Amazing work 👏 

@aharpervc I agree with all of this

> investigate if we can make a macOS job

This should be relatively straightforward and really similar to linux. The harder part will be getting the mssql-server-linux-tinytds docker image running on the MacOS executor https://support.circleci.com/hc/en-us/articles/360045029591-Can-I-use-Docker-within-the-macOS-executor-

> For example, we might benefit from having "test" be separate from "build", and then require test to succeed before build runs.

Yes, definitely something we should do but agree these are future goals. Not anything that needs to be accomplished within this PR. It would also be nice to add automated releases as part of that work.

bryanwieg:
I am working on this today. Do we want ruby 2.5 in the build matrix, as it is already past eol?

Any thoughts?

bryanwieg:
Also, as I moved on to looking at how the windows tests would work. If I read this right, CircleCI Windows 2022 image does have sql server 2019 developer pre-installed. But I can see the old appveyor appears to have had a custom image with 3 different (older) sql server instances to test against.

At the moment I can see that testing the windows port may take some consideration. At least eh Windows matrix is working vs two Ruby versions. 

I'm going to move back to restoring the Linux build. The tests there should be much easier with docker.

But I welcome any thoughts about how to set up testing on the windows side.

bryanwieg:
makes sense, thank you

aharpervc:
> If I read this right, CircleCI Windows 2022 image does have sql server 2019 developer pre-installed. But I can see the old appveyor appears to have had a custom image with 3 different (older) sql server instances to test against.

IIRC the sql server image used for testing is a mostly empty image, and is available to make running SQL Server easier. I'm not clear on how easily we can run _Linux_ docker image on a _Windows_ host in CircleCI (considering you can do this apart from CI you'd think it'd be possible, but whatever).

BTW, I made the veracross/circleci-ruby-freetds image to pre-build FreeTDS for back in the days when the source code was only available via a flaky FTP server that'd go down a lot and the git repo didn't have version tags. More recent version of FreeTDS _do_ have version tags on GitHub: https://github.com/FreeTDS/freetds/tags. Therefore rather than using that custom base image, we can probably build FreeTDS either via the rake task or by cloning the repo and switching to a supported/tested tag, or the most recent release, etc

bryanwieg:
As far as I know, it would not be possible to run a linux container on windows without hyperv support and docker desktop. Which I highly doubt is possible in circleci. MS official prebuilt SQL for Windows Containers were discontinued long ago (they were also >12GB large). Even if we made some, they would be very big.

This means, as far as what circleci does provide for windows SQL Server, it would only be SQL Server 2019. Which I think we could use for testing. But I see no way to test older versions of SQL server without installing them in the CI run, which would take TON of ci minutes.

aharpervc:
I agree, using the built-in SQL Server instance is probably the best option for now. I don't view _not_ being able to parameterize the SQL Server version on _Windows_ as a blocker here at all. I don't even think we absolutely need to parameterize that for Linux per se, and we should be able to use `mcr.microsoft.com/mssql/server:2019-latest` for that accessory image so at least it's the same between OS jobs.

Then it'll just be a matter of making sure we're able to run whatever scripts we need ahead of time to prepare that SQL Server instance for the automated tests, ie, these scripts: https://github.com/rails-sqlserver/docker-images/tree/master/mssql-server-linux-tinytds/tinytds

bryanwieg:
Sounds good. Just to check for sure tough, we are saying only SQL Server 2019 for both windows and Linux is good?

aharpervc:
> Sounds good. Just to check for sure tough, we are saying only SQL Server 2019 for both windows and Linux is good?

I think so, yes. We could probably set up additional on a per-version basis for SQL Server for Linux, but it doesn't seem critical right now. It might _become_ critical in the future, but not necessary for this PR.

Although it probably wouldn't be terribly difficult to support multiple versions. For example, if the CI jobs were `windows_mssql-latest`, `linux_mssql-2019` and `linux_mssql-2017` (with reference to the corresponding SQL Server docker image tags) that'd do a lot. Then you can imagine the level of effort to add a CI job for SQL Server 2022 when that docker image gets tagged. (Then we'll have a clear answer when people come here and post a question "hey, does TinyTDS support SQL Server 2022????")

aharpervc:
We probably won't be this lucky, but I am a bit curious if the [setup_remote_docker](https://circleci.com/docs/2.0/configuration-reference/#setupremotedocker) feature would give a Linux docker connection if run from a Windows job? 🤔

If it actually worked that way, that'd be ideal, since the Ruby test code on Windows could run SQL Server on Linux via a container, and we wouldn't need to rely on the pre-installed SQL Server for Windows instance.

(Probably won't be that lucky though....)

bryanwieg:
I'm just posting this here so it is not forgotten. I don't know know when (or if it a good idea to use these)

Windows Docker SQL Containers SQL 2012-2019.

https://hub.docker.com/r/cagrin/mssql-server-ltsc2022

bryanwieg:
@bvogelzang, regarding the tiny_tds circleci pipeline build...

I know its been a few months since I was able to resume this- I have a question please.. I'm pinging you because I see from a past mr that you may be familiar with some of the code involved in my issue.

For example: https://github.com/bryanwieg/tiny_tds/commit/10433d8fcf268f4278e8a3fef042ca176db38d39

I have linux working now in circleci (recall windows was already completed). It builds and runs the tests.

However, two tests are failing, blocking me. Both tests _work_, they just fail the assertion. Both have to do with toxiproxy tests. One fails with an error returned different than one that was expected. The other test fails expecting an error but none was thrown.

https://app.circleci.com/pipelines/github/bryanwieg/tiny_tds/72/workflows/aa995130-9fa7-40a5-8060-295e8a715c9a/jobs/91

```  
1) Failure:
With in-valid options#test_0008_raises TinyTds exception with dead connection network failure [/home/circleci/repo/test/client_test.rb:168]:
Expected: 20047
  Actual: 20003

2) Failure:
With in-valid options#test_0009_raises TinyTds exception with login timeout [/home/circleci/repo/test/client_test.rb:187]:
expected a TinyTds::Error but none happened
```

Could you please take a look? I was able to replicate the problem using local docker containers for development.

I spent considerable time backdating versions of everything to match past builds on tiny_tds where these tests passed. Such as SQL Server 2017 GA, free_tds, and other gems. But I cannot get past these assersions. I don't know if it's a build environment issue, or if something else has changed meaning the original issue's these tested for have changed and the code itself needs to be addressed.

You can use these general steps to get a dev container up to see the issue on your own machine.
1. it's all in docker-compose: `docker-compose up`
2. attach to the cimg/ruby container: `docker container exec -it tiny_tds-cimgruby-1 /bin/bash `
3. build the dependencies: `. ~/repo/test/bin/setup_cimg_dev.sh`
4. build the gem and run the tests: `sudo -E bundle exec rake`

Thanks for your time.

bryanwieg:
@aharpervc, regarding the above post- in the mean time, I have disabled those two tests and moved forward.

In the latest pipeline, both windows and linux successfully build, and linux will successfully test. Both on all Ruby versions 2.5.x to 2.7.x.

https://app.circleci.com/pipelines/github/bryanwieg/tiny_tds/74/workflows/7d2bf856-c96a-4752-8a88-1ebfd2ae5dd6

Was there anything else that should be done in this particular pull request? As I go alone, I see all sorts of things that could be done, but I'm trying to keep the scope under control so that we can at least make reasonable baby steps. Also, I could use a second set of eyes to see if anything is missing.

Thanks,
Bryan

.circleci/config.yml
```diff
@@ -0,0 +1,182 @@
...
+    steps:
+      - checkout
+
+      - run:
+          name: setup sql
+          command: |
+            curl https://packages.microsoft.com/keys/microsoft.asc | sudo apt-key add -
+            curl https://packages.microsoft.com/config/ubuntu/20.04/prod.list | sudo tee /etc/apt/sources.list.d/msprod.list
+            sudo apt-get update
+            sudo ACCEPT_EULA=Y apt-get -y install mssql-tools unixodbc-dev
```
aharpervc, line 44: It's my preference with this config to have a greater number of more specific steps (vs having fewer, bigger steps). This code seems to be installing some prerequisites, so what do you think about moving that to it's own step? Then "setup sql" can just be running the two scripts

.circleci/config.yml
```diff
@@ -0,0 +1,182 @@
...
+            /opt/mssql-tools/bin/sqlcmd -S sqlserver -U sa -P $SA_PASSWORD -i ./test/sql/db-create.sql
+            /opt/mssql-tools/bin/sqlcmd -S sqlserver -U sa -P $SA_PASSWORD -i ./test/sql/db-login.sql
+
+      - run:
+          name: install dependencies
+          command: |
+            sudo ./test/bin/install-openssl.sh
+            sudo ./test/bin/install-freetds.sh
```
aharpervc, line 52: Same comment here; it seems like this would be best as separate steps. I realize this is the equivalent of the "rake ports" step for windows, but I somewhat wish _that_ was split out too

.circleci/config.yml
```diff
@@ -0,0 +1,182 @@
...
+
+      - run:
+          name: test gem
+          command: |
+            bundle exec rake test
+
+  test_windows:
+    parameters:
+      ruby_version:
+        description: "version tag for the veracross/circleci-ruby-freetds container"
```
aharpervc, line 85: Same description as above

.circleci/config.yml
```diff
@@ -0,0 +1,182 @@
...
+        type: string
+
+    executor:
+      name: win/server-2022
+      shell: powershell.exe
+
+    environment:
+      RAKEOPT: "-rdevkit"
+      TESTOPTS: '-v'
+      MAKE: "make V=1 -j2"
```
aharpervc, line 95: Is it possible to use `'` vs `"` uniformly in this config?

.circleci/config.yml
```diff
@@ -0,0 +1,182 @@
...
+          paths:
+            - ./vendor/bundle
+          key: v1-bundle-<< parameters.ruby_version >>-{{ .Branch }}-{{ checksum "tiny_tds.gemspec" }}
+
+      - run:
+          name: build
+          no_output_timeout: 30m
+          command: |
+            $Env:PATH = "C:\\Ruby<< parameters.ruby_version >>-x64\\bin;$Env:PATH"
+            bundle exec rake ports
```
aharpervc, line 152: Is it worth splitting this out so that each dependency is a separate step?

aharpervc (commented):
Bryan, this is outstanding work, bravo 👏. I left a bunch of line comments to consider, mostly marginal improvements to reduce questions in the future. As to what else for this PR, I think we should also delete the appveyor config (and skim the readme for references, replace the build status icon with a shield link [per this doc](https://circleci.com/docs/2.0/status-badges/), etc).

Here's what I see as reasonable next steps
1. make any additional minor changes
2. remove appveyor config & references
3. get feedback from others
4. merge
5. celebrate

test/client_test.rb
```diff
@@ -154,45 +154,45 @@ class ClientTest < TinyTds::TestCase
       end
     end
 
-    it 'raises TinyTds exception with dead connection network failure' do
-      skip if ENV['CI'] && ENV['APPVEYOR_BUILD_FOLDER'] # only CI using docker
```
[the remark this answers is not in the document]
bryanwieg: that's the strange thing with those tests, I could not get them to pass assertion even outside ci, on my local machine. I hope to have input more on this. In any case, I can find a way to disable them in ci for now

.circleci/config.yml
```diff
@@ -0,0 +1,182 @@
...
+      - image: cimg/ruby:<< parameters.ruby_version >>
+        environment:
+          TESTOPTS: "-v"
+          TINYTDS_UNIT_DATASERVER: sqlserver
+          TINYTDS_UNIT_HOST: sqlserver
+          SA_PASSWORD: super01S3cUr3
+          TOXIPROXY_HOST: toxiproxy
+
+      # Service container image
+      - image: mcr.microsoft.com/mssql/server:2019-CU16-ubuntu-20.04
```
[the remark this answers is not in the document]
bryanwieg: this is just the latest build of the container. the MSSQL containers don't go by version numbers but by year + cu. I can change it to GA (from 2019), or latest. I do feel that pinning any version is better than 'latest'. I'll set to GA for now unless you have another preference.

.circleci/config.yml
bryanwieg: can you clarify this?

.circleci/config.yml
```diff
@@ -0,0 +1,182 @@
...
+        name: sqlserver
+        environment:
+          ACCEPT_EULA: Y
+          SA_PASSWORD: super01S3cUr3
+
+      - image: shopify/toxiproxy:2.1.4
```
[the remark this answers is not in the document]
bryanwieg: When I looked back, this was the version used the last time those tests worked. I believe newer versions will break the tests more. 

Is it alright if we leave toxiproxy for later? I feel it makes it clear for the next person what version was last used for those tests. And I hate to pin a version of something I am fairly confident won't work at all without updated tests. 

Otherwise, I can pin it to the latest version, but the toxiproxy stuff will all got to get addressed at some point. Let me know.

bryanwieg:
I pinned versions of some things back to the last time pipelines ran, just to get it running in a seeming consistent way as it did before. Even updating SQL seemed like a strech. I was planning to leave the real updating of dependencies to another mr.

.circleci/config.yml
bryanwieg: ohhh I see, yes I think I can do this

.circleci/config.yml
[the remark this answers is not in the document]
aharpervc: How about this: switch to the GH registry, but keep the reference to this version, and add a brief comment explaining that it's on purpose

.circleci/config.yml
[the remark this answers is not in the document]
aharpervc: Well, it _happens_ to be the latest build of the container. How about `2019-latest` or similar?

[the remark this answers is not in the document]
bryanwieg: yes, I can do that

.circleci/config.yml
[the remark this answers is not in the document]
bryanwieg: @aharpervc, does it matter what organization the final circleci pipeline is registered under? I ask since the shield committed will point to this.

bryanwieg:
@aharpervc, there is a pipeline running now.. assuming it all runs, the pr is ready for another review :)

README.md
```diff
@@ -1,11 +1,9 @@
 # TinyTDS - Simple and fast FreeTDS bindings for Ruby using DB-Library.
 
-* [![TravisCI](https://travis-ci.org/rails-sqlserver/tiny_tds.svg?branch=master)](https://travis-ci.org/rails-sqlserver/tiny_tds) - TravisCI
-* [![Build Status](https://ci.appveyor.com/api/projects/status/g2bhhbsdkx0mal55/branch/master?svg=true)](https://ci.appveyor.com/project/rails-sqlserver/tiny-tds/branch/master) - Appveyor
+* [![<ORG_NAME>](https://circleci.com/<VCS>/<ORG_NAME>/<PROJECT_NAME>.svg?style=svg)](<LINK>) - CircleCi
```
aharpervc, line 3: Don't forget to fill in this template

bryanwieg:
Hope you guys are all doing well. Just a reminder that the pr is still up whenever you have time.

bryanwieg:
crazy timing, circleci just sent out an email about a plan for open source projects:
https://circleci.com/open-source/

docker-compose.yml
```diff
@@ -20,3 +20,17 @@ services:
       - "1234:1234"
     networks:
       - main-network
+
+  cimgruby:
```
bvogelzang, line 24: Depending on what the solution ends up being I'm not sure we need this to be defined here. It's entirely appropriate to run the tests from your host machine. I could see this being useful if we end up making use of this within our circleci execution though and so am open to keeping it around.

bvogelzang (changes requested):
@bryanwieg It appears that the toxiproxy tests are failing because the tests are bypassing the proxy. The reason is that the `TINYTDS_UNIT_DATASERVER` environment variable is set. That's being used instead of the host/port combination overridden in the toxiproxy tests which results in the proxy being bypassed. Removing that environment variable setting appears to resolve issues running the tests locally.

The second part of the problem is that the networking for the containers is not preserved when running in the circleci environment. This causes inconsistencies when running locally vs. in circleci.

Per: https://circleci.com/docs/2.0/using-docker#using-multiple-docker-images
> All containers run in a common network and every exposed port will be available on localhost from a [primary container](https://circleci.com/docs/2.0/glossary#primary-container).

We're not making use of `docker-compose` within circleci which means the `1234` port on the toxiproxy image we need is not being exposed. I don't know of a way to expose this port using the docker executor. That gives us a few options on how to proceed:

1. We can use a linux machine executor and utilize `docker-compose`. We could then run the tests within those containers.
2. Use the docker executor but use remote docker by calling `setup_remote_docker`. Configure/run tests within the remote docker container and network we have set up.
3. We install toxiproxy directly in the main docker `cimg/ruby` container and start the proxy in the background once installed via `toxiproxy-server`. I haven't been able to get this to work yet but feel like I'm missing a small configuration change which would get things working.
4. Some combination of the above which I'm not thinking of...

Hopefully this can give you some things to try moving forward.

bvogelzang:
It looks like option number 3 will work. Here is the commit to get the toxiproxy tests passing: https://github.com/bvogelzang/tiny_tds/commit/b83445c3edb48c186ab0e4fc1b4bb1bc57f21508. Hope that helps!

bryanwieg:
sorry for the delay. 

thanks for replicating this. The tests connecting to the sql server container work fine. I wonder why then they do not connect to toxiproxy container. So this is helpful, I was so convinced it was the tests themselves I never tried just installing the toxyproxy package to verify that.

I'll try to dig into it more. If connecting to the sql server container works, I can not see why there is not a way to get it connecting to the toxiproxy container also. The docs say that the tests are running from the first container cimg/ruby and from there connect to the other containers.

examples:
* https://circleci.com/docs/2.0/postgres-config
* https://circleci.com/blog/setting-up-tricky-containers-in-circle-2-0-multi-image/
  * The first section _only_, to show an example of someone adding a sql container and connecting to it

I'll keep digging. I feel like I would like to use the toxiproxy container, but I may yield and give in and just install the package if it does not work.

bryanwieg:
ahhhhh, yes, good point. I'll be sure to look into that.

bryanwieg:
based on @bvogelzang insights, I was able to find the problem and get it to work in docker and using docker-compose.

To do it in circleci would require using a linux machine environment and docker-compose.

I am going to present a version of this that pr uses this method. Two of the benefits being that you could develop locally all using containers, and using the exact same build environments locally as well an in the ci pipeline.

bryanwieg:
@aharpervc, well it is ready for a look, all tests passing!

bryanwieg:
Hello @aharpervc and @bvogelzang, hope you are doing well, just wondering if you had any thoughts on this so far.

bryanwieg:
Hey @aharpervc , @bvogelzang , hope everyone is doing well! Let me know if there is anything I can do to nudge this forwards

ecentell-CPF:
I'd also like to nudge this forward as it may allow PR503 to be merged.

aharpervc:
I haven't forgotten; just buried with other things. I have a few misc questions/thoughts.

1. We need to figure out the circleci org. I'm assuming it'd be preferable for it to be named `rails-sqlserver` like the GH org here
2. There's still a lot of "manual setup" going on in these steps, such as `docker exec cimg_ruby bash -c 'sudo -E ./test/bin/install-freetds.sh'` and manually downloading/installing ruby for windows. If that's how it has to be, then so be it I guess. And we can always iterate later (the priority is definitely getting _some_ functional cross platform CI operational...)
3. Remind me why we have to do "docker exec cimg_ruby bash ..." as an approach generally at all?
4. I'd like input from any other contributors that are around cc @wpolicarpo

bryanwieg:
> 1. We need to figure out the circleci org. I'm assuming it'd be preferable for it to be named `rails-sqlserver` like the GH org here

This is like- the last thing, that I know of, we are waiting for. I don't know how it needs to be done. For example, does a maintainer need to set api keys for circleci to hook into the repo? I think circleci wants api keys

> 2. There's still a lot of "manual setup" going on in these steps, such as `docker exec cimg_ruby bash -c 'sudo -E ./test/bin/install-freetds.sh'` and manually downloading/installing ruby for windows. If that's how it has to be, then so be it I guess. And we can always iterate later (the priority is definitely getting _some_ functional cross platform CI operational...)

manually downloading and installing ruby is required for it to build in windows due to various limitations 

> 3. Remind me why we have to do "docker exec cimg_ruby bash ..." as an approach generally at all?

what you are seeing is circleci (running from the linux executer vm) hooking to into the cimg ruby container and building the environment

the reason we are using containers on top of a full linux vm is to accommodate toxiproxi and localhost networking, which required a complete native docker environment

this approach has the byproduct of allowing the entire application to be developed locally in a container (via docker-compose up) identically to how it would be built in the ci

> 4. I'd like input from any other contributors that are around cc @wpolicarpo

aharpervc:
> This is like- the last thing, that I know of, we are waiting for. I don't know how it needs to be done. For example, does a maintainer need to set api keys for circleci to hook into the repo? I think circleci wants api keys

Hmm, maybe. This is an open source repo so you wouldn't think there'd need to be any secrets. Maybe it's a matter of approval at the org level or something?

> manually downloading and installing ruby is required for it to build in windows due to various limitations

"Manual" is the question here, not the step itself. I have a vague memory that one of us tried to use scoop or chocolately and it didn't work?

> the reason we are using containers on top of a full linux vm is to accommodate toxiproxi and localhost networking, which required a complete native docker environment
> this approach has the byproduct of allowing the entire application to be developed locally in a container (via docker-compose up) identically to how it would be built in the ci

That makes sense, thanks. I wonder if it's be less confusing (or less work) to install toxiproxy some other way, rather than trying to maintain an identical setup to local development. Regardless, I don't think this question needs to block merging this PR, since if we figure out a better way in the future at least we'll have had functional CI all that time

bvogelzang:
>  I wonder if it's be less confusing (or less work) to install toxiproxy some other way

I would vote to go this route as I suggested in my comment above:

> It looks like option number 3 will work. Here is the commit to get the toxiproxy tests passing: [bvogelzang@b83445c](https://github.com/bvogelzang/tiny_tds/commit/b83445c3edb48c186ab0e4fc1b4bb1bc57f21508). Hope that helps!

Toxiproxy install/config becomes two steps in `.circleci/config.yml` instead of using the container

```yml
      - run:
          name: install toxiproxy
          command: |
            wget -O toxiproxy-2.1.4.deb https://github.com/Shopify/toxiproxy/releases/download/v2.1.4/toxiproxy_2.1.4_amd64.deb
            sudo dpkg -i toxiproxy-2.1.4.deb
      - run:
          name: start toxiproxy
          background: true
          command: |
            sudo toxiproxy-server -port 8474 -host localhost
```

While we would lose the ability for local development/tests to run as it is in CI I think this is an appropriate trade-off to reduce complexity as opposed to hiding CI configuration behind various shell scripts. Just my opinion. However, I agree this doesn't need to block merging this PR.

bryanwieg:
> "Manual" is the question here, not the step itself. I have a vague memory that one of us tried to use scoop or chocolately and it didn't work?

that is correct, quite a lot of time was spent on chocolatey and scoop. in the end those would repackage the rubydev and msys2 tools separately and install them in such a way that the dependencies such as openssl and libiconv would not build. 

finally switched to installing the native rubydev with msys2 bundle using its own installer and it 'just worked'

I understand about the containers thing, it was a suggestion. It has helped me quite a lot to be able to do all this locally because circleci will run out of minutes rather quickly if you try pushing many, many small changes to figure stuff out. but I don't develop on this project as you all do. that said, I am thankful that it can squeeze in just so that this can move forward.

what I am referring to when trying to add a repo to a project in circleci is this:

image

bryanwieg:
hello all, just wanted to check in on the tinytds circleci pr. see if we are waiting on anything from me

ecentell-CPF:
I'd also like to nudge this forward as it may allow PR503 to be merged.

ecentell-CPF:
Any update on this?

bryanwieg:
I haven’t heard anything since the last comments, unfortunately.

bryanwieg:
I am ready to keep working on it if the project is still maintained. But since it seems to have slowed down, we have had to look at forking or removing this dependency.

aharpervc:
> I am ready to keep working on it if the project is still maintained. But since it seems to have slowed down, we have had to look at forking or removing this dependency.

It's "maintained" in the sense of accepting volunteer contributions like you have done here in this PR, which benefit everyone. I don't think anyone has blocking complaints about the code itself, but there is the open question of the CircleCI config.

Regarding that...

> what I am referring to when trying to add a repo to a project in circleci is this:

Is there an option for GitHub on that list? Your screenshot there shows GitLab

bryanwieg:
thanks @aharpervc, my mistake regarding gitlab. let me check that now...

.circleci/config.yml
```diff
@@ -0,0 +1,182 @@
...
+          name: remove pre-installed ruby
+          command: |
+            Get-ChildItem -path 'C:\tools\' -filter Ruby* | Remove-Item -Force -Recurse
+
+      - run:
+          name: install ruby devkit
+          command: |
+            [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12;
+            Invoke-WebRequest -UseBasicParsing -uri "https://github.com/oneclick/rubyinstaller2/releases/download/RubyInstaller-<< parameters.ruby_version >>/rubyinstaller-devkit-<< parameters.ruby_version >>-x64.exe" -OutFile ruby-setup.exe;
+            .\ruby-setup.exe /VERYSILENT /NORESTART /DIR=C:/Ruby<< parameters.ruby_version >>-x64;
```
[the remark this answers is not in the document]
bryanwieg: it's optional for powershell.. i'll look at removing the semi colon and rerunning the pipeline to make sure its ok

bryanwieg:
i apologize, the point regarding my gitlab/github mistake pointed me down a path that ultimately answered, i think, how to do it.  

i believe once this is merged, a circleci org will automatically be created and associated with the org owners of the repo here in github- that is to say `rails-sqlserver`. the link i've supplied -i believe- should point to it's location in circleci based on the documentation.

bryanwieg:
my org checked just recently on this as a matter of fact. if we can keep this moving and updated to ruby 3.0, it will be very helpful to us and allow us to help maintain it as well. thank you for your help

aharpervc:
> i believe once this is merged, a circleci org will automatically be created and associated with the org owners of the repo here in github- that is to say `rails-sqlserver`. the link i've supplied -i believe- should point to it's location in circleci based on the documentation.

It's possible that we'll have to manually set up the project in CircleCI. I just clicked the button to (apparently?) authorize that integration for the rails-sqlserver GH org. It doesn't appear that the integration can be previewed with the new config file coming from a fork, but that should be point & click after merging this, if my understanding is correct.

bryanwieg:
thank you @aharpervc :)

what would be the next steps? do we still need an approval from @bvogelzang

aharpervc:
> is there a preference on cleaning up the commit history or commit message?

In this case the intermediate branch commits don't add anything useful to the master history, so I recommend squashing here on the branch first before merging

bryanwieg:
squashed, and circleci is running on the local branch just to check one last time

bryanwieg:
well, it does look like that tests have all passed. as far as i understand it's done. 

please let me know if there is anything left for the pr

thank you all so much for your time and help

bryanwieg:
@aharpervc , I was just curious, was there news from circleci about fixing the org and activating the pipeline?

PR #520, opened 2022-11-29
Update gem maintainers

aharpervc:
I'm proposing we update the gemspec config to reflect the more recent volunteer era. Keeping metaskills first on the list still of course.

---

<details>
Also, I need a dummy commit to see if that will cause a CircleCI job to run...
</details>


bf4:
I'm indifferent. We can have a build task which updates these if desired.

PR #521, opened 2022-11-30
ruby 3x support

bryanwieg:


bryanwieg:
@aharpervc , if you have a moment, just let me know if this is headed in a good direction. seems ruby 3.0 is passing with little change needed

as a matter of note, a bit off topic perhaps, for the short term:
i have already discovered that ruby 3.1 introduces some sort of breaking change to the gem.

all tests fail as unable to connect to sql server. and, even worse, the gem causes the 2017-ga version of mssql to crash upon attempting to connect. mssql cli will even return error "TCP Provider: Error code 0x2746" until the sql service is restarted. although later versions of mssql do not hard crash when the gem attempts to connect, the gem is still unable to successfully connect and pass any tests.

i only spent a short time on it. the resolution may reside in another dependency that needs to be updated, and i figured for now it may be good to just focus on ruby 3.0 since it seems rather straight forward.

please let me know any observations or thoughts you have

test/bin/setup_volume_permissions.sh
```diff
@@ -1,10 +1,21 @@
...
+# grants circlci docker uid and gid required write permissions to project file system
 
 set -x
 
-sudo groupadd -g 3434 circleci_tinytds
-sudo usermod -a -G circleci_tinytds $USER
-sudo useradd circleci_tinytds -u 3434 -g 3434
-sudo usermod -a -G circleci_tinytds circleci_tinytds
-sudo chgrp -R circleci_tinytds .
+if $(dpkg --compare-versions $ruby_version lt 3.1)
```
aharpervc, line 6: Hm, all of this is a bit confusing. Is it to prepare for Ruby 3.1?

docker-compose.yml
```diff
@@ -5,7 +5,7 @@ networks:
 
 services:
   mssql:
-    image: metaskills/mssql-server-linux-tinytds:2017-GA
+    image: mcr.microsoft.com/mssql/server:2022-RTM-ubuntu-20.04
```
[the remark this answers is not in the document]
bryanwieg: actually, i wondered if you had any thoughts about this

mssql 2022 came out this month. do you have a preference for 2022 or to stay on 2019-latest?

 i suppose as long as 2022-latest passes everything, i don't see a reason not to test on it.

test/bin/setup_volume_permissions.sh
bryanwieg: it was from the 3.1 work, yes. circleci changed the permissions in their container. i will stash this change for a future commit

andyundso:
hello @bryanwieg / @aharpervc, I wanted to ask what's the state with this pull request as I'm currently blocked on one project that uses Windows to upgrade to Ruby 3.

Given what's documented here, I could also cherry-pick the necessary changes for Ruby 3 and create a new PR to get a timely release (or pre-release). Then we could continue with implementing the proposed CI improvements and start on the Ruby 3.1 / 3.2 changes. what do you think?

bryanwieg:
it may be best for you to open another pr. i've had to divide my attention to multiple projects here.

As a side note, i want to report that i found the issue with ruby 3.1 and the tests. it looks like it has to do with tls. setting the build option for freetds to not use openssl and instead use gnutls `--with-gnutls` seems to help with the issue.

in any case, it could potentially be a few months before i can resume work on these pr's

PR #524, opened 2023-01-07
Support Ruby 3.0

andyundso:
This PR adds support for Ruby 3.0. It needed a couple of changes, including a few that are technically for Ruby 2.5 support.

* Delete the extra minitest entry in the Gemfile. As explained in the commit, the gemspec locked minitest to v5.14.0, but the version from the Gemfile had higher priority, which meant it installed 5.17.0 which isn't compatible anymore with Ruby 2.5.
* Lock bundler version on CI to 2.3.16, as this is the last release that supports Ruby 2.5.
* I tried to build the Windows native gem by running `bundle exec rake gem:windows`, but this failed with a Rake error that it is unable to find this task. I updated the Rake task in this project to match the example from the `rake-compiler-dock` docs, which made the build start.
* I got a very long error message when trying to compile with OpenSSL v1.1.1d, so I updated it to v1.1.1s.
* And at last, compile the ports and the gem itself against Ruby 3.0.

I tested the changes against a project of mine. It's a "traditional" Rails v6.1 application using 6.1.2.1. I ran our CI with a version of the gem build based on my changes and everything is green with the following combinations:

* Ruby 2.7, x64-mingw32, SQL Server Express 2017
* Ruby 3.0, x64-mingw32, SQL Server Express 2017
* Ruby 2.7, x86_64-linux, FreeTDS v1.2.20, SQL Server Express 2017
* Ruby 2.7, x86_64-linux, FreeTDS v1.2.20, SQL Server Express 2019
* Ruby 3.0, x86_64-linux, FreeTDS v1.2.20, SQL Server Express 2017
* Ruby 3.0, x86_64-linux, FreeTDS v1.2.20, SQL Server Express 2019

Credits to @bryanwieg for the initial work in #521.


bryanwieg:
@andyundso this looks really nice!

@aharpervc will probably be the one to review it.

tasks/native_gem.rake
```diff
@@ -5,10 +5,10 @@ task 'gem:windows' => ['ports:cross'] do
...
   # make sure to install our bundle
-  build = ['bundle']
+  sh "bundle package --all"   # Avoid repeated downloads of gems by using gem files from the host.
 
   # and finally build the native gem
-  build << 'rake cross native gem RUBY_CC_VERSION=2.7.0:2.6.0:2.5.0:2.4.0 CFLAGS="-Wall" MAKE="make -j`nproc`"'
-
-  RakeCompilerDock.sh build.join(' && ')
+  GEM_PLATFORM_HOSTS.keys.each do |plat|
+    RakeCompilerDock.sh "bundle --local && RUBY_CC_VERSION='2.7.0:2.6.0:2.5.0:2.4.0' CFLAGS='-Wall' MAKE='make -j`nproc`' rake native:#{plat} gem", platform: plat
```
aharpervc: 1. This command switched from `rake cross native gem` to `rake native:{plat}`. Why does this need to change?
2. I see we have Ruby 2.4 on the cross compile list. Given the other commit about bundler (etc), and that Ruby 2.4 is well out of support, what do you think about dropping it? That'd mean that our current release is the last one tested on 2.4. We have the same questions about 2.5 and 2.6, although those ought to continue to work for the time being, so I don't think we also need to drop those right now.

ext/tiny_tds/extconsts.rb
```diff
@@ -2,7 +2,7 @@
 ICONV_VERSION = ENV['TINYTDS_ICONV_VERSION'] || "1.15"
 ICONV_SOURCE_URI = "http://ftp.gnu.org/pub/gnu/libiconv/libiconv-#{ICONV_VERSION}.tar.gz"
 
-OPENSSL_VERSION = ENV['TINYTDS_OPENSSL_VERSION'] || '1.1.1d'
+OPENSSL_VERSION = ENV['TINYTDS_OPENSSL_VERSION'] || '1.1.1s'
```
aharpervc, line 5: I assume this is fine. Are the 1.1.1 release variants cross compatible with each other? Is there any downside to moving as far forward as the latest `l` (L) release?

tasks/native_gem.rake
```diff
@@ -9,6 +9,6 @@ task 'gem:windows' => ['ports:cross'] do
 
   # and finally build the native gem
   GEM_PLATFORM_HOSTS.keys.each do |plat|
-    RakeCompilerDock.sh "bundle --local && RUBY_CC_VERSION='2.7.0:2.6.0:2.5.0:2.4.0' CFLAGS='-Wall' MAKE='make -j`nproc`' rake native:#{plat} gem", platform: plat
+    RakeCompilerDock.sh "bundle --local && RUBY_CC_VERSION=#{RUBY_CC_VERSION} CFLAGS='-Wall' MAKE='make -j`nproc`' rake native:#{plat} gem", platform: plat
```
aharpervc, line 12: Would you mind moving this to a separate commit, such that adding 3.0 to the list _only_ adds 3.0 and _no_ other change?

aharpervc:
This is cool... very exciting

ext/tiny_tds/extconsts.rb
aae42: FYI, `l` doesn't refer to latest, they're just going through the alphabet on releases for `1.1.1`, and [`s` is the most recent](https://www.openssl.org/news/cl111.txt) at the moment

andreasnef:
Would be great to get this as I assume many (as us) are planning moving to v3 soon.

orgads:
Please add also 3.1 and 3.2.

orgads:
Thank you!

andyundso:
Replaced with #530.

Issue #515, opened 2022-04-12
Ruby 3.2 warning: undefining the allocator of T_DATA class TinyTds::Result


jeremyevans:
When running tiny_tds on Ruby 3.2.0-preview 1, you get a warning when first accessing the `TinyTds::Result` class:

```
warning: undefining the allocator of T_DATA class TinyTds::Result
```

You can probably avoid this by adding `rb_undef_alloc_func(cTinyTdsResult);` after https://github.com/rails-sqlserver/tiny_tds/blob/e3e79a030271891b11b6e7676382fe31f389b099/ext/tiny_tds/result.c#L586

For details, see https://bugs.ruby-lang.org/issues/18007 and https://ruby-doc.org/core-3.1.1/doc/extension_rdoc.html#label-C+struct+to+Ruby+object

stillhart:
Wow, @jeremyevans, you were early :-). I get this warning now with the official release of Ruby 3.2.0, so it's still relevant.

PR #486, opened 2021-01-19
Prepare for 2.1.4

aharpervc:
#### todo

- [x] public prerelease incorporating these with assistance from @bvogelzang & co
  - [x] https://github.com/rails-sqlserver/tiny_tds/pull/475
  - [x] https://github.com/rails-sqlserver/tiny_tds/pull/481
  - [ ] ~https://github.com/rails-sqlserver/tiny_tds/pull/469~ (later)
- [x] get feedback on this prerelease and/or those PR's
- [x] merge the PR's
- [x] update the changelog here
- [x] update to 2.1.4 final here
- [ ] merge this PR
- [ ] public release

PR #533, opened 2023-05-10
Support Ruby 3.1 and 3.2

andyundso:
This PR introduces support for Ruby 3.1 and Ruby 3.2 to `tiny_tds`. Couple of obstacles:

The GCC compiler for x64 UCRT within the `rake-compiler-dock` environment is named the same as the one for x64 mingw32, but they yield completely different binaries. This introduces an issue with the port mechanic, where we used the `host` property from `RbConfig` to determine the compiler and where to save the ports. I introduced an additional parameter for passing the Ruby architecture to the ports compilation tasks. The Ruby architecture is used to still have a way to differ between the ports for UCRT and Mingw32 while they share the name for the GCC compiler.

The cross-compile job for Windows now runs in parallel for each platform. This was because of another effect: When trying to install the compiled gems for Windows to get the compiled libraries, it raised `Unable to find spec for #<Gem::NameTuple tiny_tds, 2.1.5, x64-unknown>`, but only for 2.7 and below. Turned out that when installing a gem locally, RubyGems parsed all gem specifications for any file named `*.gem` in the current directory. And since older versions of Rubygems don't know ucrt, it raised `x64-unknown`. I found it out by essentially running `byebug /my/ruby/installation/bin/gem install tiny_tds`, opening the source code of RubyGems to find out where it got the specs from and setting a breakpoint using an absolute path again.

The parallisation added a couple of other issues with CircleCI, namely:

- The ports are cached for each Ruby version. I had a couple of faulty CI runs where compiling `tiny_tds` only worked for some Ruby versions when the ports were restored from cache. I assume there are differences in regards to the container environment in the different images.
- As the precompiled gems are saved into "the workspace", the files need to be unique. As each job also added the non-native version into the workspace, it resulted in a conflict. I solved this by removing the non-native gem so only the precompiled versions are saved into the workspace.

On CircleCI, the Linux tests initially failed for Ruby 3.1 and 3.2 when they tried to compile OpenSSL (see [here](https://app.circleci.com/pipelines/github/simplificator/tiny_tds/127/workflows/b3da0d4b-1655-4778-9a1e-79fe4e7593ad/jobs/813)). I did not bother finding out why, but instead used the ports mechanic from the gem to get the necessary libraries to build `tiny_tds`.

Note that this PR does not address #515.


ecentell-CPF:
Any hopes to get this merged soon?

smtadaka:
Hi, When can we expect this PR to be merged?

ecentell-CPF:
Hoping to get this PR some attention for a merge

andyundso:
since a couple of comments have been posted, @aharpervc would you mind to have a look at my PR?

pere73:
It would be great, if tiny_tds would support at least Ruby 3.1 on Windows.

tiny_tds.gemspec
```diff
@@ -22,8 +22,8 @@ Gem::Specification.new do |s|
   s.metadata['msys2_mingw_dependencies'] = 'freetds'
   s.add_development_dependency 'mini_portile2', '~> 2.5.0'
   s.add_development_dependency 'rake', '~> 13.0.0'
-  s.add_development_dependency 'rake-compiler', '~> 1.1.0'
-  s.add_development_dependency 'rake-compiler-dock', '~> 1.1.0'
+  s.add_development_dependency 'rake-compiler', '~> 1.2.0'
+  s.add_development_dependency 'rake-compiler-dock', '~> 1.3.0'
```
aharpervc, line 26: I wonder if we even need the patch level specifier on these... maybe it'd be best as `'rake-compiler', '~> 1.2'` so the minor version is floating?

Rakefile
```diff
@@ -6,11 +6,24 @@ require 'rake/extensiontask'
...
+ruby_cc_mingw32_versions = "3.0.0:2.7.0:2.6.0:2.5.0:2.4.0".freeze
+
 GEM_PLATFORM_HOSTS = {
-  'x86-mingw32' => 'i686-w64-mingw32',
-  'x64-mingw32' => 'x86_64-w64-mingw32'
+  'x64-mingw32' => {
+    host: 'x86_64-w64-mingw32',
+    ruby_versions: ruby_cc_mingw32_versions
+  },
+  'x64-mingw-ucrt' => {
```
aharpervc, line 22: Looks like this commit is also introducing a ucrt based build, can you update the commit message with an additional explanation on this?

test/gem_test.rb
```diff
@@ -9,11 +9,11 @@ class GemTest < MiniTest::Spec
 
     # We're going to muck with some system globals so lets make sure
     # they get set back later
-    original_host = RbConfig::CONFIG['host']
+    original_platform = RbConfig::CONFIG["arch"]
```
aharpervc, line 12: Prefer single quote string literals to match elsewhere in the file

ext/tiny_tds/extconf.rb
```diff
@@ -21,9 +21,11 @@ def do_help
 do_help if arg_config('--help')
 
 # Make sure to check the ports path for the configured host
-host = RbConfig::CONFIG['host']
+architecture = RbConfig::CONFIG['arch']
+architecture = "x86-mingw32" if architecture == "i386-mingw32"
+
 project_dir = File.expand_path("../../..", __FILE__)
-freetds_ports_dir = File.join(project_dir, 'ports', host, 'freetds', FREETDS_VERSION)
+freetds_ports_dir = File.join(project_dir, 'ports', architecture, 'freetds', FREETDS_VERSION)
```
aharpervc, line 28: I don't completely understand the swap from "host" to "architecture", other than that I generally agree the latter makes sense. Is switching this matching rake compiler patterns, or is it something bespoke for this project's config?

.circleci/config.yml
```diff
@@ -197,35 +197,26 @@ jobs:
...
   cross_compile_gem:
-    machine:
-      image: ubuntu-2004:current
+    parameters:
+      platform:
+        description: "Platform to compile the gem resources"
+        type: string
+
+    docker:
+      - image: "ghcr.io/rake-compiler/rake-compiler-dock-image:1.3.0-mri-<< parameters.platform >>"
```
aharpervc, line 218: Switching the base image here is a little surprising, can you share more why this is needed? Also, rather than pinning to 1.3.0, is there a tag for the latest release?

.circleci/config.yml
```diff
@@ -39,30 +39,42 @@ jobs:
...
+          keys:
+            - ports-<< parameters.ruby_version >>-{{ checksum ".ports_versions" }}
+            - ports-<< parameters.ruby_version >>-
 
       - run:
-          name: bundle install gems
+          name: compile ports and build gem
           command: |
-            docker exec cimg_ruby bash -c 'bundle install'
+            docker exec cimg_ruby bash -c 'bundle exec rake ports && bundle exec rake build'
```
aharpervc, line 60: I recommend keeping "rake ports" and "rake build" as separate CI steps unless they absolutely need to be combined

.circleci/config.yml
```diff
@@ -247,6 +247,12 @@ jobs:
             mkdir -p artifacts-<< parameters.platform >>/gems
             mv pkg/*.gem artifacts-<< parameters.platform >>/gems
 
+      - run:
+          name: Remove non-native gem to avoid conflict in workspace
```
aharpervc, line 251: Does this mean there's unexpected shared state between the CI jobs? Addressing that might be a better approach than "delete everything that's not me" 🤔

aharpervc (approved):
This generally seems fine, although I left a few line comments to consider.

Also, I'd like to have the CI jobs run to validate the PR, but I don't see that they have. "Build forked pull requests" is turned on in the project settings. I see the jobs have run in your fork, though, so that's probably good enough: 
image

ext/tiny_tds/extconf.rb
andyundso: not really on us, but rather on rake-compiler-dock.

I edited the description, so it's hopefully more clear.

.circleci/config.yml
andyundso: > Switching the base image here is a little surprising, can you share more why this is needed?

It's been a while since I implemented this. But basically, we previously had this one Ubuntu job which compiled the ports for `x86-mingw32`, `x64-mingw32`. Now with this PR, we need to introduce another platform with `x64-mingw-ucrt`.

To speed to process up, I instead opt to run a separate job for each architecture. I do not recall how much time it saves, but I assume we talk about ~10 minutes per pipeline run.

> Also, rather than pinning to 1.3.0, is there a tag for the latest release?

Looking into [GitHub packages](https://github.com/rake-compiler/rake-compiler-dock/pkgs/container/rake-compiler-dock-image), there is a snapshot version available, which might be unstable. Then 1.3.0 is already the latest release.

I think it makes sense to pin this version to align it with the `rake-compiler-dock` gem.

.circleci/config.yml
andyundso: It is a rather undesired side-effect of `gem build`: Each build yields its non-native version of the gem, on which CircleCI reports a conflict of resources. I did some research and there seems to be not a way to surpress the generation of the non-native gem.

Rakefile
andyundso: I added a reference to the commit in [cae55bf](https://github.com/rails-sqlserver/tiny_tds/pull/533/commits/cae55bf54932f63be4175dab9b3137acb3dbb037).

test/gem_test.rb
andyundso: addressed in [cb98ec1](https://github.com/rails-sqlserver/tiny_tds/pull/533/commits/cb98ec137aa9b7dd20a8fcff78890b40ea682a48).

tiny_tds.gemspec
andyundso: I removed the minor version constraint for `rake-compiler` in [801f16c](https://github.com/rails-sqlserver/tiny_tds/pull/533/commits/801f16ccf5685eff6817e559858e920e44fd948d). I kept it for `rake-compiler-dock` to keep it aligned with the Docker image, as stated in https://github.com/rails-sqlserver/tiny_tds/pull/533#discussion_r1336300203.

.circleci/config.yml
andyundso: addressed in [69a6dbd](https://github.com/rails-sqlserver/tiny_tds/pull/533/commits/69a6dbd0f747bf193dbf95085ee0b984aae24af1).

andyundso:
@aharpervc addressed your feedback, mind to have a second look?

aharpervc:
> @aharpervc addressed your feedback, mind to have a second look?

Seems fine. And the CircleCI integration started working for this PR, which is nice. Something I noticed is that despite having a step to upload artifacts, there isn't any files actually saved. Eg, should be a .gem file here: https://app.circleci.com/pipelines/github/simplificator/tiny_tds/137/workflows/5a2df471-ebdc-4a0c-9fdc-c87af53afb20/jobs/984/artifacts

Having that functional _is_ a goal for this project so that folks can test gem's for PR's without building locally. I suppose it doesn't need to block this PR if there's capacity to investigate soon as a followup

aharpervc (approved):

andyundso:
> > @aharpervc addressed your feedback, mind to have a second look?
> 
> Seems fine. And the CircleCI integration started working for this PR, which is nice. Something I noticed is that despite having a step to upload artifacts, there isn't any files actually saved. Eg, should be a .gem file here: https://app.circleci.com/pipelines/github/simplificator/tiny_tds/137/workflows/5a2df471-ebdc-4a0c-9fdc-c87af53afb20/jobs/984/artifacts
> 
> Having that functional _is_ a goal for this project so that folks can test gem's for PR's without building locally. I suppose it doesn't need to block this PR if there's capacity to investigate soon as a followup

Will do so!

ecentell-CPF:
Thanks for getting this merged!  Can you tell me will there be a version increment or a related push to rubygems.org?

aharpervc:
Yep, if nobody else has made the PR yet, go for it. I recommend this general pattern as a template: https://github.com/rails-sqlserver/tiny_tds/pull/486 (plus a commit that appends `.pre` to the version string, as a temporary commit to drop before merging).

However...

> [re: assets from CI] Will do so!

I recommend basing the version prep PR on 👆, so someone with rubygems access can use the assets from CI for the rubygems release, so we can also validate that CI can produce release-able assets.

PR #541, opened 2023-09-28
Fix path to artifacts

andyundso:
As I parallelized the cross-compilation process, the gems were moved into another folder dedicated to their target platform during the build. This change was not reflected when storing the artifacts.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -213,6 +213,19 @@ jobs:
       - store_test_results:
           path: test/reports
 
+      - run:
+          name: Rename gem to a consistent name to store artifact
+          command: |
+            $Env:PATH = "C:\\Ruby<< parameters.ruby_version >>-x64\\bin;$Env:PATH"
+            $rubyArchitecture = (ruby -e 'puts RUBY_PLATFORM').Trim()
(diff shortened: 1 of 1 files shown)


aharpervc:
I don't see anything on the artifacts tab here, should that be showing something now? https://app.circleci.com/pipelines/github/andyundso/tiny_tds/2/workflows/e920c589-ee26-4afd-ba5c-fd10bd257931/jobs/35/artifacts

aharpervc:
I see. That sort of makes sense, although I'm concerned about accessibility. How I checked was I scrolled to the bottom of this PR, clicked "show all checks", clicked into the details for "windows-2.4", clicked artifacts, and didn't see anything.

I think it'd be great if repo visitors had quick access to the fat gems for whichever version of ruby they're using, quickly from each PR.

Thoughts?

andyundso:
generally agree with you.

I would keep the artifacts upload on the `precompile` step, as the x86 version is not used in any test. But additionally, upload the artifact from each test, so people can get the correct version for their system, as suggest by you.

will take care of this tomorrow and amend the PR.

aharpervc (approved):
Sweet, LGTM

PR #542, opened 2023-09-28
Mark `alloc` as undefined for `TinyTds::Result`

### LINKED ISSUE #515: Ruby 3.2 warning: undefining the allocator of T_DATA class TinyTds::Result

andyundso:
From my understanding, you only need to define an `alloc` function if you want to define a C struct and store values on it. For example, `TinyTds::Client` defines such an `allocate` function to wrap its client. On `TinyTds::Result`, it does not seem we do have a struct and therefore does not need an `alloc` function.

Closes #515

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * Add Ruby 3.0 to the cross compile list
 * Fix segfault when asking if client was dead after closing it. Fixes #519.
+* Mark `alloc` function as undefined on `TinyTds::Result`. Fixes #515.
 
 ## 2.1.5
 
--- a/ext/tiny_tds/result.c
+++ b/ext/tiny_tds/result.c
@@ -584,6 +584,7 @@ void init_tinytds_result() {
   cDate = rb_const_get(rb_cObject, rb_intern("Date"));
   /* Define TinyTds::Result */
   cTinyTdsResult = rb_define_class_under(mTinyTds, "Result", rb_cObject);
+  rb_undef_alloc_func(cTinyTdsResult);
   /* Define TinyTds::Result Public Methods */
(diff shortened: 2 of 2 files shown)


aharpervc:
I assume that the warning from https://github.com/rails-sqlserver/tiny_tds/issues/515 is no longer showing with this?

andyundso:
Exactly. You can compare the output on the Ruby 3.2 job on the current `master` and this branch, and you see that the warning no longer shows on the very top of the output.

* master: https://app.circleci.com/pipelines/github/andyundso/tiny_tds/1/workflows/39fa47e7-2353-4a83-801a-c0b2d44f06e0/jobs/26/parallel-runs/0/steps/0-115
* feature branch: https://app.circleci.com/pipelines/github/andyundso/tiny_tds/3/workflows/90ab46ac-8cc5-4a2f-8c94-f84053c26b63/jobs/50/parallel-runs/0/steps/0-115

Michoels:
Looks great!
When will this be released?

taylorthurlow:
> Looks great! When will this be released?

This looks like it went out with [v2.1.6](https://github.com/rails-sqlserver/tiny_tds/releases/tag/v2.1.6) in early December.

PR #531, opened 2023-04-21
Look for freetds in /opt/homebrew

caleb:
Homebrew recently changed the directory it installs into from /usr/local to /opt/homebrew. This change allows users to install tiny_tds on macOS.

PR #545, opened 2023-12-03
Add `/opt/homebrew` to search path on Apple Silicon

andyundso:
This PR contains the same changes as in #531, but adds a job on CircleCI to verify the installation on Macs with Apple Silicon. I also made it conditional, so we do not look in `/opt/homebrew` by default.


andyundso:
as discussed with @aharpervc via mail, this is fix we like to include in the next release, so I'm going ahead and merge the branch.

Issue #552, opened 2024-02-18
tiny_tds v3


andyundso:
I have a couple of ideas for a potential v3 release of `tiny_tds`. However, I did not work that much on code so far and also not sure if this is the direction that the project wants to take, so consider this issue a proposal.

Potential content of the release:

* Drop support for older Ruby versions
    * I personally would drop everything until Ruby 3.1 (v3 wouldn't be released before end of March anyhow, and Ruby 3.0 is EOL after March 2024)
    * Otherwise, Rails 6.1 is still supported, which requires Ruby 2.5 at least, so dropping support for all versions below 2.5 could also be an option.
* Drop the 32-bit Windows build.
    * It is currently untested on CI, which is a potential risk.
    * v2.1.6 had 150 downloads for it (1200 for the 64-bit version), v2.1.7 95 so far (415 for the 64-bit version), so I would say demand for it is quite low.
* Bump the dependencies for the precompiled Windows version.
    * If possible, we could move to OpenSSL v3 when updating FreeTDS.
 * Drop support for older MSSQL versions
     * Microsoft has two support models: Mainstream and [Extended](https://learn.microsoft.com/en-us/sql/sql-server/end-of-support/sql-server-end-of-support-overview?view=sql-server-ver16#lifecycle-dates).
     * I think we should drop support for versions where extended support has been discontinued.
* Instead of returning `false` when something went wrong with the query, throw an exception.
    * Throwing an error is much more clear than `false` and v3 is the best time to do so.

Again, these are just a couple of ideas. I also read through #511 which suggest a new implementation based on the .NET client, but I personally would not like installing a .NET runtime for my Ruby project.

What do you think @aharpervc?

aharpervc:
Sounds good, especially changing the return false thing... absurd. I'd also like to consider refactoring how escape is called, so that a live connection object isn't required (maybe a class method, or maybe decoupling init & connect).

andyundso:
Hi @aharpervc, all the changes I would like to have for a version 3 are on main. Any objections to release this version somewhere this week?

aharpervc:
> Hi @aharpervc, all the changes I would like to have for a version 3 are on main. Any objections to release this version somewhere this week?

Go for it 👍

PR #553, opened 2024-03-12
Drop support for Ruby < 2.7

andyundso:
I assume the `activerecord-sql-adapter` is the source of most of our Gem installations. Rails 7.0 requires Ruby 2.7 at minimum. I think this a good base line for now. It would still allow people with older Rubys to upgrade to the eventual v3.0 release, but we do no longer support ancient Ruby versions (like 2.4).

Related to #552.


aharpervc (approved):
LGTM

PR #556, opened 2024-04-23
Switch to GitHub Actions

andyundso:
This PR switches the pipeline for tiny_tds from CircleCI to GitHub Actions.

Functionality-wise, they should be equivalent, although I made a couple of additions:

- The ports are compiled in one step for all the Linux jobs, similar to how we do this for Windows.
- Test summary is displayed for each job, e.g. [here](https://github.com/andyundso/tiny_tds/actions/runs/8806805186). 
- The built gem versions for Windows can also be downloaded from that page.
- I also test the installation of the gem on Mac with M1 processors. This was something I added once to CircleCI for #545, only to realise it is only available for paid organizations.

Open questions / points from my side:

- Currently, the installation on Windows for MSSQL 2017 is broken, so I commented it out in the matrix and reported an issue upstream (https://github.com/potatoqualitee/mssqlsuite/issues/27). Worst case we could re-activate the solution we used for CircleCI, as I did not find another action that can install multiple versions of SQL server and on different platforms. I could also fork (or re-implemented) a similar solution on my own time.
- `docker-composee.yml` currently uses the CircleCI Ruby image. I like having this compose file around for a quick start into tiny tds development, but I would suggest to switch the official Ruby images from Docker Hub.
- There are a bunch of scripts we no longer would need (like `setup_cimgruby_dev`).


.github/workflows/ci.yml
```diff
@@ -0,0 +1,443 @@
...
+      - uses: actions/checkout@v4
+
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby-version }}
+
+      - name: Download precompiled gem
+        uses: actions/download-artifact@v4
+        with:
+          name: gem-x64-mingw32
```
aharpervc, line 71: I guess we can't/won't parameterize this for the install task, because we're matching the VM?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,443 @@
...
+        run: |
+          $rubyArchitecture = (ruby -e "puts RbConfig::CONFIG['arch']").Trim()
+          $gemVersion = (Get-Content VERSION).Trim()
+          $gemToInstall = "./tiny_tds-$gemVersion-$rubyArchitecture.gem"
+          
+          Write-Host "Looking to install $gemToInstall"
+          gem install --local "$gemToInstall"
+          
+          ruby -e "require 'tiny_tds'; puts TinyTds::Gem.root_path"
+          exit $LASTEXITCODE
```
aharpervc, line 84: I recommend splitting this to it's own step, for better clarity on failure

.github/workflows/ci.yml
```diff
@@ -0,0 +1,443 @@
...
+      matrix:
+        mssql-version:
+          #- 2017
+          - 2019
+          - 2022
+        ruby-version:
+          - 2.7
+          - '3.0'
+
+    name: Test on Windows (MingW)
```
aharpervc, line 100: I think for PR check clarity, we'll benefit from seeing the ruby version (and maybe sql server version) directly in the job name. Probably same comment for all jobs as applicable. Also, I'm big fan of consistency across jobs, so maybe something like `test-$os (ruby $ruby_version)`, `build- ...`, etc. Looking at the list here: https://github.com/andyundso/tiny_tds/actions/runs/8806805186/job/24172426029

.github/workflows/ci.yml
```diff
@@ -0,0 +1,443 @@
...
+          # Restore precompiled code
+          $source = (Resolve-Path ".\tmp\gems\tiny_tds-$gemVersion-$rubyArchitecture\lib\tiny_tds").Path
+          $destination = (Resolve-Path ".\lib\tiny_tds").Path
+          Get-ChildItem $source -Recurse -Exclude "*.rb" | Copy-Item -Destination {Join-Path $destination $_.FullName.Substring($source.length)}
+
+          # Restore ports
+          Copy-Item -Path ".\tmp\gems\tiny_tds-$gemVersion-$rubyArchitecture\ports" -Destination "." -Recurse
+
+      - name: Setup MSSQL
+        uses: potatoqualitee/mssqlsuite@v1.7
```
aharpervc, line 134: Are we able to use service containers here instead?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,443 @@
...
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install FreeTDS
+        run: brew install freetds
+        shell: bash
+
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby-version }}
```
aharpervc, line 432: Do we need bundler-cache: true on all these? Why the difference?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,443 @@
...
+
+  install-windows-mingw:
+    needs:
+      - cross-compile
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby-version:
+          - 2.7
+          - '3.0'
```
aharpervc, line 57: If any value has to be quoted, I recommend establishing the pattern that they'll all always be quoted, now & going forward, throughout

.github/workflows/ci.yml
```diff
@@ -0,0 +1,443 @@
...
+          exit $LASTEXITCODE
+
+  compile-native-ports:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: 3.3
```
aharpervc, line 322: Does this need to be a matrix var?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,443 @@
...
+          - 3.1
+          - 3.2
+          - 3.3
+
+    name: Install on Mac OS (M1)
+    runs-on: macos-14
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install FreeTDS
```
aharpervc, line 426: Possible to cache this step?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,443 @@
...
+          sa-password: c0MplicatedP@ssword
+          show-log: true
+
+      - name: Setup MSSQL database
+        shell: pwsh
+        run: |
+          & sqlcmd -S localhost -U sa -P "c0MplicatedP@ssword" -i ./test/sql/db-create.sql
+          & sqlcmd -S localhost -U sa -P "c0MplicatedP@ssword" -i ./test/sql/db-login.sql
+
+      - name: Install toxiproxy-server
```
aharpervc, line 147: Possible to cache this?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,443 @@
...
+        shell: pwsh
+        run: bundle exec rake test
+        env:
+          TOXIPROXY_HOST: "localhost"
+
+      - name: Test Summary
+        uses: test-summary/action@v2
+        with:
+          paths: "test/reports/TEST-*.xml"
+        if: always()
```
aharpervc, line 163: I'm curious, how do test failures show up here?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,443 @@
+name: Build and test gem
```
aharpervc, line 1: Also with the question(s) of how checks show up in the PR UI, this might be better as `name: ci`

.github/workflows/ci.yml
```diff
@@ -0,0 +1,443 @@
...
+
+on:
+  push:
```
aharpervc, line 4: We _might_ want to limit push to master, keeping the rest of the on's as you have it. Another way to say it, I don't think pushes to branches w/o PR's needs to run this CI workflow...

.github/workflows/ci.yml
```diff
@@ -0,0 +1,443 @@
...
+    runs-on: ubuntu-latest
+    container:
+      image: "ghcr.io/rake-compiler/rake-compiler-dock-image:1.4.0-mri-${{ matrix.platform }}"
+    steps:
+      - uses: actions/checkout@v4
+      - run: git config --global --add safe.directory /__w/tiny_tds/tiny_tds # shrug
+      - name: Install gems
+        shell: bash
+        run: bundle install
+      - name: Write used versions into file
```
aharpervc, line 26: My fake undiagnosed OCD demands more newlines for this job, same as you have the others 😆

aharpervc (commented):
Awesome! Mostly just offering pedantic comments/reflections...

.github/workflows/ci.yml
bf4: fwiw, there's also this https://github.com/ankane/setup-sqlserver/

bf4 (commented):
given the complexity of this PR, it might be better to begin adding a GitHub ci action and then testing extending it before removing the circle ci config

bf4:
quick and dirty brainless importer didn't work.  but now that there's an action on main with workflow dispatch, you can test it in a PR 

example failures https://github.com/rails-sqlserver/tiny_tds/actions/runs/8888886154

.github/workflows/ci.yml
andyundso: technically we are matching the Ruby version, everything below 3.0 is mingw32, 3.1 and newer is ucrt.

we probably could make a parameter out of it, or combine it with `matrix`, but to keep it simple I would suggest to spell it out explicitly for now.

.github/workflows/ci.yml
andyundso: addressed with dc3b1e47.

.github/workflows/ci.yml
andyundso: I added a suggestion with 0b125119.

.github/workflows/ci.yml
andyundso: Linux service containers are not available on [Windows for GitHub Actions](https://github.com/orgs/community/discussions/25491). On Linux, the referenced action will use containers.

re: the action from ankane: would also work, also misses support for MSSQL 2017.

.github/workflows/ci.yml
andyundso: ... maybe?

I think it's just an executable for Windows, which we could cache. Chocolatey is just comfortable to get it running quickly. Installation also only takes about 1min, compared to MSSQL (~4min) and Ruby (~3min) it is "fast".

.github/workflows/ci.yml
andyundso: I made a branch with a test that fails intentionally. here is the result: https://github.com/andyundso/tiny_tds/actions/runs/8900358242

here is also a screenshot:

image

.github/workflows/ci.yml
andyundso: we only need to compile the ports once for all the Linux jobs, so no matrix (aka parallel jobs) needed.

.github/workflows/ci.yml
andyundso: apparently possible, but looks messy:
https://stackoverflow.com/a/65056232

.github/workflows/ci.yml
andyundso: I think was just an oversight. I added it now to all jobs: 49c557d0.

andyundso (commented):
@aharpervc I think I addressed all your feedback.

I would adjust the development docker compose in a second PR, to not make this one even more bloated. :)

aharpervc (approved):
LGTM

bf4 (approved):
Great to see this green. 

If it's ok to merge I'm happy to merge it.

Then I'll remove the not-actually-working circleci->gh-actions import I added whilst trying to be helpful (.github/workflows/test.yml and .github/actions/install-ruby-windows/action.yml) and if y'all have admin to remove the circleci check we can remove that else put in a no-op config till we can get Ken to remove it

PR #560, opened 2024-05-13
Raise error if dbsqlsend fails

### LINKED ISSUE #464: Result from a dead connection is false. Should it throw an exception?
## Before submitting an issue please check these first!

* On Windows? If so, do you need devkit for your ruby install?
* Using Ubuntu? If so, you may have forgotten to install FreeTDS first.
* Are you using FreeTDS 0.95.80 or later? Check `$ tsql -C` to find out.
* If not, please update then uninstall the TinyTDS gem and re-install it.
* Have you made sure to [enable SQL Server authentication](http://bit.ly/1Kw3set)?
* Doing work with threads and the raw client? Use the ConnectionPool gem?

If none of these help. Please fill out the following:

## Environment

**Operating System**

```
lsb_release -a; uname -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.4 LTS
Release:	18.04
Codename:	bionic
Linux app04 4.15.0-91-generic #92-Ubuntu SMP Fri Feb 28 11:09:48 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
```

```
tsql -C
[TinyTds][v1.3.0][tsql]: /usr/local/bin/tsql
Compile-time settings (established with the "configure" script)
                            Version: freetds v1.1.24
             freetds.conf directory: /usr/local/etc
     MS db-lib source compatibility: no
        Sybase binary compatibility: no
                      Thread safety: yes
                      iconv library: yes
                        TDS version: auto
                              iODBC: no
                           unixodbc: yes
              SSPI "trusted" logins: no
                           Kerberos: no
                            OpenSSL: no
                             GnuTLS: no
                               MARS: yes
```

## Description

Previously posted this on the [sequel gem](https://github.com/jeremyevans/sequel/issues/1695). It looks like there's an expectation for the TinyTDS client to raise an exception when the connection is dead, not return false.

The following shows that the client is returning false:

```ruby
require 'sequel'
require 'logger'
DB = Sequel.connect(ENV['DATABASE_URL'])
DB.extension(:connection_validator)
DB.loggers << Logger.new($stdout)
DB[:table].first # Successful
# Externally kill the connection to the DB
DB[:table].first # Reports the error below
```

```
/home/aex/fno-interface-stage/vendor/bundle/ruby/2.7.0/gems/sequel-5.32.0/lib/sequel/adapters/tinytds.rb:73: warning: TinyTds: dbsqlsend() returned FAIL.
Traceback (most recent call last):
       16: from /home/aex/.rvm/gems/ruby-2.7.0/bin/irb:23:in `<main>'
       15: from /home/aex/.rvm/gems/ruby-2.7.0/bin/irb:23:in `load'
       14: from /home/aex/.rvm/rubies/ruby-2.7.0/lib/ruby/gems/2.7.0/gems/irb-1.2.1/exe/irb:11:in `<top (required)>'
       13: from (irb):9
       12: from /home/aex/fno-interface-stage/vendor/bundle/ruby/2.7.0/gems/sequel-5.32.0/lib/sequel/dataset/actions.rb:208:in `first'
       11: from /home/aex/fno-interface-stage/vendor/bundle/ruby/2.7.0/gems/sequel-5.32.0/lib/sequel/dataset/actions.rb:693:in `single_record'
       10: from /home/aex/fno-interface-stage/vendor/bundle/ruby/2.7.0/gems/sequel-5.32.0/lib/sequel/dataset/actions.rb:705:in `single_record!'
        9: from /home/aex/fno-interface-stage/vendor/bundle/ruby/2.7.0/gems/sequel-5.32.0/lib/sequel/dataset/actions.rb:950:in `with_sql_first'
        8: from /home/aex/fno-interface-stage/vendor/bundle/ruby/2.7.0/gems/sequel-5.32.0/lib/sequel/dataset/actions.rb:942:in `with_sql_each'
        7: from /home/aex/fno-interface-stage/vendor/bundle/ruby/2.7.0/gems/sequel-5.32.0/lib/sequel/adapters/tinytds.rb:214:in `fetch_rows'
        6: from /home/aex/fno-interface-stage/vendor/bundle/ruby/2.7.0/gems/sequel-5.32.0/lib/sequel/dataset/actions.rb:1089:in `execute'
        5: from /home/aex/fno-interface-stage/vendor/bundle/ruby/2.7.0/gems/sequel-5.32.0/lib/sequel/adapters/tinytds.rb:46:in `execute'
        4: from /home/aex/fno-interface-stage/vendor/bundle/ruby/2.7.0/gems/sequel-5.32.0/lib/sequel/database/connecting.rb:270:in `synchronize'
        3: from /home/aex/fno-interface-stage/vendor/bundle/ruby/2.7.0/gems/sequel-5.32.0/lib/sequel/connection_pool/threaded.rb:92:in `hold'
        2: from /home/aex/fno-interface-stage/vendor/bundle/ruby/2.7.0/gems/sequel-5.32.0/lib/sequel/adapters/tinytds.rb:77:in `block in execute'
        1: from /home/aex/fno-interface-stage/vendor/bundle/ruby/2.7.0/gems/sequel-5.32.0/lib/sequel/adapters/tinytds.rb:217:in `block in fetch_rows'
NoMethodError (undefined method `fields' for false:FalseClass)
```

andyundso:
Closes #464
Relates to #552

With this PR, `tiny_tds` will raise an error in case `dbsqlsend` will return `FAIL` instead of returning false. It is essentially the code from #469 / @wpolicarpo.

I spent an evening trying to build a test that would yield this error. I came up with a solution that worked on my machine, but only sometimes on CI.

```ruby
it 'raises error when query cannot be sent' do
  client = new_connection
  assert_client_works(client)

  thread1 = Thread.new do
    client.execute("SELECT 1 as [one]").do
  end

  thread2 = Thread.new do
    assert_equal false, client.execute("SELECT 1 as [one]")
  end

  thread1.join
  thread2.join
end
```

I assume, when the result are read in thread1 and Ruby decides to switch to thread2, our FreeTDS wrapper is in an invalid state, which can trigger the issue. Likely, because they both try to access the DBPROCESS object, which [should not be allowed](https://www.freetds.org/faq.html#thread.safe).

I also tried various suggestions to kill the session or just restart the database server, none of them worked for me to trigger the error. I personally think we can still get this code in even if it has no tests, as it can avoid getting into weird state with tiny_tds.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 * Drop support for Ruby < 2.7
 * Drop support for SQL Server < 2017
 * Drop support for FreeTDS < 1.0
+* Raise error if FreeTDS is unable to sent command buffer to the server
 
 ## 2.1.7
 * Add Ruby 3.3 to the cross compile list
--- a/ext/tiny_tds/client.c
+++ b/ext/tiny_tds/client.c
@@ -299,8 +299,7 @@ static VALUE rb_tinytds_execute(VALUE self, VALUE sql) {
   REQUIRE_OPEN_CLIENT(cwrap);
   dbcmd(cwrap->client, StringValueCStr(sql));
   if (dbsqlsend(cwrap->client) == FAIL) {
-    rb_warn("TinyTds: dbsqlsend() returned FAIL.\n");
-    return Qfalse;
+    rb_raise(cTinyTdsError, "failed dbsqlsend() function");
   }
   cwrap->userdata->dbsql_sent = 1;
   result = rb_tinytds_new_result_obj(cwrap);


aharpervc:
Idea, try adding `sleep 1` in thread1 after the statement, eg

```ruby
thread1 = Thread.new do
  client.execute("SELECT 1 as [one]").do
  sleep 1
end
```

andyundso:
I pushed the test now @aharpervc. As you can see, it sometimes works, sometimes not. So it really depends on a race-conditon between the two threads.

aharpervc:
Yeah, unfortunate, I thought we could force an order of operations with the sleep. It seemed to be reliable for me locally with that technique. Is there a way to kill the client without threads?

andyundso:
Hi @aharpervc

I experimented this evening again and I did not find any way to write a test that could simulate a `FAIL` from `dbsqlsend`. The primary case where this could happen is when the SQL server goes down between us checking if it is available with `REQUIRE_OPEN_CLIENT` and then running `dbsqlsend`, as illustrated below.

```c
REQUIRE_OPEN_CLIENT(cwrap);
dbcmd(cwrap->client, StringValueCStr(sql));
if (dbsqlsend(cwrap->client) == FAIL) {
  rb_raise(cTinyTdsError, "failed dbsqlsend() function");
}
```

I assume the reason why the thread-based test sometimes work is because the `userdata` or the command buffer of FreeTDS is manipulated at the correct moment, and so the test fails. which, like we saw, is not reproducable really.

I now removed the commit with the test and suggest to merge this change as it is.

aharpervc (approved):

PR #561, opened 2024-07-12
Add `bigdecimal` as a dependency

andyundso:
This is to address the following warning on Ruby 3.3:

```
/home/apf/dev/rails-sqlserver/tiny_tds/lib/tiny_tds.rb:3: warning: bigdecimal was loaded from the standard library, but will no longer be part of the default gems since Ruby 3.4.0. Add bigdecimal to your Gemfile or gemspec.
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,7 +82,7 @@ jobs:
           $gemToInstall = "./tiny_tds-$gemVersion-$rubyArchitecture.gem"
 
           Write-Host "Looking to install $gemToInstall"
-          gem install --local "$gemToInstall"
+          gem install "$gemToInstall"
 
       - name: Test if TinyTDS loads
         shell: pwsh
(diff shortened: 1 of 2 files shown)


andyundso:
will check somewhere this weekend what is needed to get the Windows pipeline running.

andyundso:
The tests on Windows work now again: Instead of using `gem install` into a temporary directory as we did before, I now use `gem unpack` to extract the gem.

andyundso:
> Also, I don't understand why we need a ci step to install bigdecimal manually if it's listed in the gemspec. Why wouldn't the normal gem install tiny_tds process bring in that new dependency?

Good point, `gem install` should actually do that. I checked and most of our `gem install` commands used `--local`, which restricted the gem to download anything from the internet. This is removed now in the latest commit and the additional step to install bigdecimal is no longer required.

> This seems weird, can you explain more what that's about?

I think `gem unpack` is a more elegant solution to solve what we want to do, which is to extract the compiled files from the built gem.

Issue #539, opened 2023-09-27
Crashes (SIGABRT) Under Test Automation


clintmiller:
## Environment

**Operating System**

```
ProductName:		macOS
ProductVersion:		13.4.1
ProductVersionExtra:	(c)
BuildVersion:		22F770820d
```

**TinyTDS Version and Information**

```
[TinyTds][v2.1.5][tsql]: /usr/local/bin/tsql
Compile-time settings (established with the "configure" script)
                            Version: freetds v1.3.20
             freetds.conf directory: /usr/local/etc
     MS db-lib source compatibility: no
        Sybase binary compatibility: yes
                      Thread safety: yes
                      iconv library: yes
                        TDS version: 7.3
                              iODBC: no
                           unixodbc: yes
              SSPI "trusted" logins: no
                           Kerberos: yes
                            OpenSSL: yes
                             GnuTLS: no
                               MARS: yes
```

**FreeTDS Version**

```
freetds: 1.3.20
```

## Description

In the course of trying to upgrade an application from Ruby 2.5/Rails 5.2.x to Ruby 3.1/Rails 6.1.x, I have noticed a significant increase in automated test failures (we're using `rspec-rails`). These failures seem take a couple of forms:

#### Terminal output:
```
<big long memory region dump trimmed>

Abort trap: 6
```
#### OS Crash Report:
 ```
Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	    0x7ff8092e61f2 __pthread_kill + 10
1   libsystem_pthread.dylib       	    0x7ff80931dee6 pthread_kill + 263
2   libsystem_c.dylib             	    0x7ff809244b45 abort + 123
3   libruby.3.1.dylib             	       0x106e845e9 die + 9
4   libruby.3.1.dylib             	       0x106e84810 rb_bug_for_fatal_signal + 544
5   libruby.3.1.dylib             	       0x106fb475b sigsegv + 91
6   libsystem_platform.dylib      	    0x7ff80934b5ed _sigtramp + 29
7   ???                           	               0x0 ???
8   tiny_tds.bundle               	       0x10c34543e nogvl_dbsqlok + 88 (result.c:130) [inlined]
9   tiny_tds.bundle               	       0x10c34543e rb_tinytds_result_ok_helper + 109 (result.c:179) [inlined]
[...]

PR #563, opened 2024-10-16
Refactor `do`, `insert` and `each`

andyundso:
Relates to #539

FreeTDS enforces a strict sequence of how queries are sent and results are retrieved. To summarize it very briefly:

* Open a connection using `dbopen`.
* Prepare a command buffer using `dbcmd`.
* Send it to the server using `dbsqlsend`.
* Acknowledge that the server acknowledge using `dbsqlok`.
* Gather metadata using `dbdbresults`.
* Fetch each row using `dbnextrow`.
* Cancel the running results (`dbcancel`) or close the client using `dbclose`.

You do not have to serialize / bind each retrieved result, but you still need to fetch them. This is mentioned [in the sample code](https://www.freetds.org/userguide/samplecode.html) as well:

> DB-Library doesn't insist every column — or even any column — be bound or otherwise retrieved into the application's variables. There is, however, one absolutely crucial, inflexible, unalterable requirement: the application must process all rows produced by the query.

The current implementation of `do` and `insert` perform the entire sequence. However, `each` works different: Since `each` lazy-loads each result from the server, it does not peform the entire sequence.

As mentioned on #539, the error "Assertion `conn->in_net_tds == NULL'" likely occured when people did not call `cancel` on the result object, as the query is technically still pending from FreeTDS point-of-view.

With this PR, in `#execute`, I check if we have to still acknowledge and fetch results prior to running another query, essentially doing the job of `cancel`  automatically for the user.

This is not my initially proposed solution on #539, in fact, I coded out the entire solution there. Ultimately, I am still unhappy with the side-effects that lazy loading the results have, but it is still a nice feature after all.


andyundso:
Replaced with https://github.com/rails-sqlserver/tiny_tds/pull/595.

PR #566, opened 2024-10-21
Pin CI to Ubuntu 22.04

andyundso:
Microsoft SQL Server 2017 will not work on Ubuntu 24.04.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
           - "x86-mingw32"
           - "x64-mingw-ucrt"
     name: cross-compile-windows
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     container:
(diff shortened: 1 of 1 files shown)


aharpervc:
For the `n`th time, remind me why we're unable to run tests on the stock container images?

andyundso:
we are using the Stock container images, but something with the SQL Server 2017 image is broken on Ubuntu 24.04 (I assume the old 18.04 Kernel of SQL Server 2017 image does not work with newer distros).

aidanharan:
It may be related to https://github.com/microsoft/mssql-docker/issues/899

andyundso:
we only install on Windows, as there we do not have container images available. on Linux, we use containers (see e.g. [here](https://github.com/rails-sqlserver/tiny_tds/actions/runs/11446915491/job/31847282989), just need to open the part which says "Setup MSSQL" and there you see the `docker pull` logs).

aharpervc:
Ah, I see potatoqualitee/mssqlsuite is running the linux image as you said, when possible

PR #568, opened 2024-12-15
Bump version to v3.0.0

### LINKED ISSUE #552: tiny_tds v3

andyundso:
Closes #552 

This PR bumps the version for the `tiny_tds` gem to v3.0.0, which I will release later today.


aharpervc:
👏

PR #573, opened 2024-12-25
Add Ruby 3.4 to the cross compile list

andyundso:
This required an update of `rake-compiler-dock`, as well as an update of `minitest` (since Ruby 3.4 no longer ships `mutex_m`, which our old version of `minitest` was using).

Currently pipelines fail because Ruby 3.4 is not yet available on Windows. But I saw that the CI on Linux still uses RC1 of Ruby 3.4.


Rakefile
```diff
@@ -7,7 +7,7 @@ require_relative './ext/tiny_tds/extconsts'
 
 SPEC = Gem::Specification.load(File.expand_path('../tiny_tds.gemspec', __FILE__))
 
-ruby_cc_ucrt_versions = "3.3.0:3.2.0:3.1.0".freeze
+ruby_cc_ucrt_versions = "3.4.0:3.3.5:3.2.0:3.1.0".freeze
```
aharpervc, line 10: Since 3.3.0 became 3.3.5, do we also want to update to 3.1.6 & 3.2.6?

tiny_tds.gemspec
```diff
@@ -24,8 +24,8 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'mini_portile2', '~> 2.5.0'
   s.add_development_dependency 'rake', '~> 13.0.0'
   s.add_development_dependency 'rake-compiler', '~> 1.2'
-  s.add_development_dependency 'rake-compiler-dock', '~> 1.4.0'
-  s.add_development_dependency 'minitest', '~> 5.14.0'
+  s.add_development_dependency 'rake-compiler-dock', '~> 1.7.0'
+  s.add_development_dependency 'minitest', '~> 5'
```
aharpervc, line 28: I think this syntax would allow updating major versions, which we probably wouldn't expect. I think `~> 5.25` would be safer

tiny_tds.gemspec
```diff
@@ -24,8 +24,8 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'mini_portile2', '~> 2.5.0'
   s.add_development_dependency 'rake', '~> 13.0.0'
   s.add_development_dependency 'rake-compiler', '~> 1.2'
-  s.add_development_dependency 'rake-compiler-dock', '~> 1.4.0'
```
aharpervc, line 27: What changed here?

aharpervc (approved):

Rakefile
andyundso: The change to 3.3.5 comes from the v1.6.0 release notes of `rake-compiler-dock`:
https://github.com/rake-compiler/rake-compiler-dock/releases/tag/v1.6.0

looking at how `nokogiri` configured their cross-compilation list, it appears to not be necessary to update any other version:
https://github.com/sparklemotion/nokogiri/blob/dadec3409556be85aecf9eff3857b883cacbe563/.cross_rubies

tiny_tds.gemspec
andyundso: * v1.5: Support for compiling gems against musl (Alpine's "glibc" library)
* v1.6: Base images are now based on Ubuntu, therefore some libraries were upgraded.
* v1.7: Support for compiling gems against Ruby 3.4

aharpervc: The release notes [here](https://github.com/rake-compiler/rake-compiler-dock/releases/tag/v1.6.0) say the minimum glibc version was incremented; does that mean _our_ minimum version of linux/glibc is also bumped for our linux releases? or is that just for the build system for windows?

andyundso: that is just for the build system on Windows.

andyundso:
I see there are a bunch of compile warnings for the Ruby 3.4 jobs (e.g. [here](https://github.com/rails-sqlserver/tiny_tds/actions/runs/12544272016/job/34995686796?pr=573)). Will merge this PR and look at those tomorrow before releasing a new version. Just want to avoid that the installation process of `tiny_tds` becomes noisy.

andyundso:
I now implemented this whole `TypedData` stuff as part of my refactoring branch:
https://github.com/rails-sqlserver/tiny_tds/pull/571/commits/f7ce48174ce1cb4f17b510d9ae523809c244ed19

In our current code structure, `GET_RESULT_WRAPPER` is in the header file for `result.c`, and I did not find out how I would properly setup this `rb_data_type_t` constant within  the header file including the references to the `free` and `mark` function. But `tiny_tds` works fine with Ruby v3.4, so I think there is no urgency for now to fix this.

PR #592, opened 2025-08-26
Bump version to v3.3.0

andyundso:
It would be good to release a small thing in-between. Likely the next release will be end of December with Ruby 3.5 support.


aharpervc:
LGTM

PR #593, opened 2025-09-03
Trivial typo fixes

justinclift:
Some typos jumped out at me while skimming. :wink:

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ Optionally, Microsoft has done a great job writing [an article](https://learn.mi
 
 ## Data Types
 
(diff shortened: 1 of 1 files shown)


justinclift:
Also noticed what kind of seems like mismatching TDS version default information in the README (bitrot?), but I didn't investigate further:

> Below is a list of the data types we support when using the 7.4 TDS protocol version.

> TDS version. Defaults to "7.3".

If the default is now something other than 7.3, then the above line might need to be updated. :smile:

andyundso:
you're correct, this should be 7.3 instead of 7.4. if you have the chance to correct this in your PR as well, that would be nice.

justinclift:
Sure, done. :smile:

Note that the follow up typo fix was done as a separate commit, so this PR will probably need squashing if/when merged. :smile:

andyundso (approved):

PR #609, opened 2026-09-16
Add Client#ping(timeout:) for FreeTDS-bounded liveness checks

emailrhoads:
## Summary
- Adds `TinyTds::Client#ping(timeout:)` that temporarily sets FreeTDS `DBSETTIME`, runs `SELECT 1`, restores the previous timeout on success, and returns `false` on timeout/failure so callers can discard the handle.
- Adds `query_timeout` / `query_timeout=` (shared helper also used at connect) so a live client can change query timeout without reconnecting.
- Does **not** change passive `active?` / `dead?` (`dbdead`) — explicit `#ping` keeps the opt-in clear for adapters/apps.

This is meant as the driver-level answer to the checkout-liveness question in [activerecord-sqlserver-adapter#1396](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/issues/1396): idle **TCP RST** (TCP reset — peer or middlebox abruptly closes the socket while the client still believes the session is alive) plus passive FreeTDS `dbdead`. Ruby `Timeout` around `execute` does not reliably interrupt TinyTDS/C I/O; the short bound has to be FreeTDS-level. Policy for when to call `#ping` (e.g. adapter `verify!` / `active?`) can live in the adapter — see follow-up [activerecord-sqlserver-adapter#1412](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1412).

## Test plan
- [ ] `bundle exec rake compile`
- [ ] `test/client_test.rb`: `query_timeout` from connect; `query_timeout=` bounds WaitFor; `#ping` returns true and restores timeout; `#ping` returns false when round-trip exceeds ping timeout; ArgumentError / closed client cases


emailrhoads:
Tracking discussion / attention request on the adapter side: https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/issues/1413 (also links this PR and adapter #1412).

